PR #14, opened 2023-08-07
tensorboard functionality

TarekAbouChakra:
Key changes:

- Tensorboard Class (tblogger): I've added a new tensorboard class, tblogger, which integrates  into neps. This integration allows us to take advantage of TensorBoard's visualization.
- Config Data Extraction: I've integrated the extraction of configuration data from both metahyper and the neps API. This step proves incredibly useful for parallel runs.
- Dependency Update: A new dependency has been introduced to support TensorBoard integration in poetry.
- Example Usage: To facilitate easy adoption, I've included a new example under neps_examples/convenience that demonstrates how to run neps with the tblogger class.

Drawback Consideration:

One potential drawback of the tblogger integration is the creation of multiple tfevent files. This was unaviodable to ensure compatibility with parallel runs.


TarekAbouChakra:
Changes:

- Removing logging gradients and leaving the user's preference by manual calculation and logging through scalars.
- Fixing the example file, making it easier to read and follow.

Remaining issues and TODO's:

- The issue of generating multiple files throughout the run.
- Finalizing the documentation.

src/neps/plot/tensorboard_eval.py
```diff
@@ -0,0 +1,522 @@
...
+                                }
+                                Default is None.
+
+        """
+        tblogger.current_epoch = current_epoch
+        tblogger.loss = loss
+        tblogger.scalar_accuracy_mode = scalar_accuracy_mode
+        tblogger.hparam_accuracy_mode = hparam_accuracy_mode
+
+        if not tblogger.disable_logging:
```
eddiebergman, line 496: Another point of the double negative. Could also flatten this quite a bit with early return

```python
if not tblogger.enabled:
    tblogger.logger_bool = False
    return
    
# Logging code
```

src/neps/plot/tensorboard_eval.py
```diff
@@ -0,0 +1,522 @@
...
+        exp, ssi, sei = hparams(hparam_dict, updated_metric)
+
+        self.file_writer.add_summary(exp)
+        self.file_writer.add_summary(ssi)
+        self.file_writer.add_summary(sei)
+        for k, v in updated_metric.items():
+            self.add_scalar(tag=k, scalar_value=v, global_step=global_step)
+
+
+class tblogger:
```
eddiebergman, line 43: After reviewing the rest of this, some of the parameters here could use some documentation as things like `logger.logger_init_bool` and `logger.logger_bool` aren't very clear from usage why they are needed. You might come up with a better name by documenting it ;)

It also seems like `logger.logger_bool` is 1-1 correlated with `disable_logging`, is `.logger_bool` needed?

You can add documentation for these kind of variables in this way. I think NePs uses `mkdocs-material` which means this will render nicely for the class if documentation every gets up there.

```python
class tblogger:
    x: T
    """This x does blah."""
    
    y: K
    """This y is for booping the booper.
    
    !!! warning "In multi-worker setup"
    
        Not all tblogger classes across workers will boop the booper
        and this variable may vary between workers.
    """
```

src/neps/plot/tensorboard_eval.py
```diff
@@ -0,0 +1,522 @@
...
+    - add_hparams: Overrides the base method to log hyperparameters and metrics with better formatting.
+    """
+
+    def add_hparams(self, hparam_dict, metric_dict, global_step):
+        if not isinstance(hparam_dict, dict) or not isinstance(metric_dict, dict):
+            raise TypeError("hparam_dict and metric_dict should be dictionary.")
+        updated_metric = {}
+        for key, value in metric_dict.items():
+            updated_key = "Summary" + "/" + key
+            updated_metric[updated_key] = value
```
eddiebergman, line 33:
```python
updated_metric = {f"Summary/{key}": val for key, val in metric_dict.items()}
```

eddiebergman (commented):
Despite the number of comments, I really like the addition! Most of the comments are just about how to improve the code to make it more maintainable in the future and just to share how I read the code as going through it from the perspective of a potential maintainer who is seeing this for the first time and may have to fix a bug or add something new, and what would help me.

Also I'm a nitpicker, sorry, feel free to ignore where you disagree!

TarekAbouChakra:
Resolved the comments in the last PR concerning:
- Code maintainability and cleaning
- Couple more warnings and error checks
- Better comments

One new feature:
- Added a tblogger.enable() to remove the double negative problem

src/neps/plot/tensorboard_eval.py
```diff
@@ -0,0 +1,522 @@
...
+
+        tblogger.config = config
+        tblogger.config_id = config_id
+        tblogger.config_working_directory = config_working_directory
+        tblogger.config_previous_directory = config_previous_directory
+        tblogger.optim_path = optim_path
+
+    @staticmethod
+    def _initialize_writers():
+        if not tblogger.config_writer:
```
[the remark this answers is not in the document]
TarekAbouChakra: Thanks for the advice, I think it should be more readable now, and NO code is running when it's not supposed to due to early return.

src/neps/plot/tensorboard_eval.py
```diff
@@ -0,0 +1,522 @@
...
+            # If the writer is still not assgined
+            optim_config_path = tblogger.optim_path / "results"
+            if tblogger.config_previous_directory is not None:
+                # If a previous directory is available (Now the search is done for higher fidelity but logging is
+                # saved on the previous directory)
+                tblogger.fidelity_mode = True
+                while not tblogger.config_writer:
```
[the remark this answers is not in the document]
TarekAbouChakra: The occurance of this is rare to non-existent. However, added to the new push an else which raises an error and exits the loop.

pyproject.toml
```diff
@@ -70,6 +70,10 @@ more-itertools = "^9.0.0"
 portalocker = "^2.6.0"
 seaborn = "^0.12.1"
 pyyaml = "^6.0"
+tensorboard = [
+    {version = "^2.11", python = "<3.8"},
```
Neeratyoy, line 74: @eddiebergman do we still need to support 3.7?

neps_examples/convenience/neps_tblogger_tutorial.py
```diff
@@ -0,0 +1,373 @@
...
+import torch.nn as nn
+import torch.nn.functional as F
+import torchvision
+from torch.optim import lr_scheduler
+from torch.utils.data.dataloader import DataLoader
+from torch.utils.data.sampler import SubsetRandomSampler
+from torchvision.transforms import transforms
+
+import neps
+from neps.plot.tensorboard_eval import tblogger
```
Neeratyoy, line 64: what happens if I import and have forgotten to do `tblogger.add_scalar(...)` and have also not done `tblogger.disable()`? 
Or does a new import of blogger in a local scope resets the state?

Can we have something like `tblogger.record_status=True/False` which the user can print to verify and debug?
I am thinking along the lines of `torch.cuda.is_available()`. 

Also, looking at the example, we should perhaps write somewhere that by default the flag to record is set to True.

pyproject.toml
eddiebergman: Probably not, we would likely want it to be `">=3.8`

neps_examples/convenience/neps_tblogger_tutorial.py
TarekAbouChakra: tblogger now has tblogger.get_status() this returns true when logger is enabled and false when it is disabled to keep track to tblogger's state.

pyproject.toml
TarekAbouChakra: I've maintained Python 3.7 support due to library dependencies. Until the library drops 3.7 support, removing it here would lead to errors for NePS users on Python 3.7.

TarekAbouChakra:
Mostly minor cleanups.

1 new addition:
- tblogger.get_status() this returns true when logger is enabled and false when it is disabled to keep track of tblogger's state.

pyproject.toml
eddiebergman: Can we at least change it to `">=3.7"` then?

eddiebergman: Important to not that officially even Python have stopped developing with 3.7 in mind by it's [end-of-life date](https://devguide.python.org/versions/)

Neeratyoy (approved):

Neeratyoy:
Could we format the example/tutorial code to follow the styling standards better, such that, it `looks` better?

TarekAbouChakra:
Changes only for the tblogger example format:
- Maximum of 80 characters per line.
- BO now does only 5 random samples before sampling from acquistion function.
- Better comments

PR #18, opened 2023-09-13
API Design Update Including YAMLs

TarekAbouChakra:
API Design Changes:

- Default Searchers as YAML Files: Users manually select NePS YAML files at `neps.run`.
- Automatic Searcher Assignment: Users can skip selection, and NEPS assigns a searcher based on the search space.
- Custom Searcher YAML Files: Users can create and link their custom searcher YAML files using `searcher_path` in `neps.run`.
- Prioritizing `kwargs`: Users can still add `kwargs` to `neps.run`, which take precedence over YAML configurations.
- Limitation on Changing `kwargs`: Changing `kwargs` is allowed only when the user specifies a searcher.

Extras:

- Replicated Searchers in `__init__` having SearcherMapping: Except for `cost_cooling` searcher.
- README File: Added detailed instructions.
- `SearchConfigs` Class: Provides insights about default NEPS searchers (algorithm, searcher names, kwargs...).

Minor Issue:

- Initially, `sample_default_first` was intended to be `True` when using priors, but it failed when having `neps.FunctionParameter` in the search space. It should now be set manually.


TarekAbouChakra:
In the latest push:
- Fixed some minor bugs.
- Added a new 'optimizer_info.yaml' file that saves essential information in the optimization directory such as the optimizer's name, algorithm, whether it's user-defined (a boolean), and whether any arguments were accepted (another boolean).
- Created test cases to validate optimizer's proper functioning. (TODO: still need to add a test case for when the searcher is designed by the user and a searcher_path is passed)

TarekAbouChakra:
Latest push:
- Creating a new marker for the yaml_api test case
- Testing only done locally for the yaml_api

tests/test_neps_api/examples_test_api.py
```diff
@@ -0,0 +1,82 @@
...
+)
+
+# Case 4: Choosing hyperband
+neps.run(
+    run_pipeline=run_pipeline,
+    pipeline_space=pipeline_space_fidelity,
+    root_directory="hyperband_neps_decided",
+    max_evaluations_total=1,
+    eta=2,
+)
```
Neeratyoy, line 82: Do we have a neps utility that given a folder can return a status whether runs have started, failed, ended, etc.? Maybe that is for a future PR.

Nevertheless, can we assert and check somehow whether the runs were successful?
Asserting if the kwargs were correctly updated or the right algorithm was called. Simple, sanity checks.

Neeratyoy, line 82: Also, not sure if we need these extra cases unless we actually want to test these optimizers specifically.
Even if so, why not the other optimizers then?
If I interpret the test correctly, Case 1 should suffice.

src/neps/utils/common.py
```diff
@@ -1,9 +1,35 @@
...
+    Returns:
+        str: The content of the YAML file.
+    """
+    folder_path = "optimizers/default_searchers"
+    script_directory = os.path.dirname(os.path.abspath(__file__))
+    parent_directory = os.path.join(script_directory, os.pardir)
+    resource_path = os.path.join(parent_directory, folder_path, f"{searcher}.yaml")
+
+    if not os.path.exists(resource_path):
+        raise FileNotFoundError(f"Searcher '{searcher}' does not exit.")
```
Neeratyoy, line 27: Can we pretty-print the list of optimizers available too?

A representative example,
```bash
ERROR!
Searcher not in:
asha
asha_prior
...
priorband
```

src/neps/optimizers/info.py
```diff
@@ -0,0 +1,108 @@
...
+import yaml
+
+
+class SearcherConfigs:
+    """
+    This class provides methods to access default configuration details
+    for NePS optimizers.
+    """
+
+    @staticmethod
```
Neeratyoy, line 14: Given the nature of the intended use of this class, would we prefer `@classmethod` or `@staticmethod` here?

tests/test_neps_api/examples_test_api.py
TarekAbouChakra: Working on status thing for a future push as we discussed.

Asserting is now included in the metahyper, checks the algorithm and simply if any kwargs were updates.

src/neps/utils/common.py
TarekAbouChakra: Done !

src/neps/optimizers/info.py
TarekAbouChakra: At the moment I left it as static class for its easy use-case. We can discuss further changes on this class.

Neeratyoy (approved):

PR #21, opened 2023-11-02
Tabular space support

Neeratyoy:


src/neps/optimizers/bayesian_optimization/acquisition_samplers/freeze_thaw_sampler.py
```diff
@@ -67,22 +89,47 @@ def _sample_new(
...
 
-        new_configs = self._sample_new(index_from=self.observations.next_config_id(), n=n)
-
-        # Set fidelity for observed configs
+        if self.is_tabular:
+            _n = n if n is not None else self.SAMPLES_TO_DRAW
+            _partial_ids = set([conf["id"].value for conf in partial_configs])
+            _all_ids = set(self.pipeline_space.custom_grid_table.index.values)
+            # accounting for unseen configs only
+            _n = max(_n, len(_all_ids - _partial_ids))
```
karibbov, line 109:
```suggestion
            _n = min(_n, len(_all_ids - _partial_ids))
```
In the case this is `max` next line will throw an error when `_n>len(_all_ids - _partial_ids)`

src/neps/optimizers/bayesian_optimization/acquisition_functions/mf_ei.py
```diff
@@ -2,41 +2,98 @@
...
+                # IMPORTANT to set the fidelity at which EI will be calculated only for
+                # the partial configs that have been observed already
+                target_fidelity = config.fidelity.value + self.b_step
+                config.fidelity.value = min(
+                    target_fidelity, config.fidelity.upper
+                )  # to respect the bounded fidelity
+            else:
+                config.fidelity.value = target_fidelity
+
+            if np.isclose(target_fidelity, config.fidelity.value):
```
karibbov, line 86: Moving this check into the above `if` statement will reduce the `numpy` comparisons when the `SAMPLES_TO_DRAW` is high

src/neps/optimizers/bayesian_optimization/acquisition_functions/mf_ei.py
```diff
@@ -46,13 +103,83 @@ def preprocess(self, x: Iterable) -> Tuple[Iterable, Iterable]:
...
+    
+    def preprocess_deep_gp(self, x: Iterable) -> Tuple[Iterable, Iterable]:
+        x, inc_list = self.preprocess(x)
+        x_lcs = []
+        for idx in x.index:
+            if idx in self.observations.df.index.levels[0]:
+                budget_level = max(0, self.get_budget_level(x[idx]) - 1)
+                lc = self.observations.extract_learning_curve(
+                    idx, budget_level
+                )
```
karibbov, line 119:
```suggestion
                budget_level = self.get_budget_level(x[idx])
                lc = self.observations.extract_learning_curve(
                    idx, budget_level
                )
```
Since `- 1` will be moved into the `extract_learning_curve` function

src/neps/optimizers/bayesian_optimization/acquisition_samplers/freeze_thaw_sampler.py
```diff
@@ -67,22 +89,47 @@ def _sample_new(
...
+            _new_configs = np.random.choice(list(_all_ids - _partial_ids), size=_n, replace=False)
+            new_configs = [self.pipeline_space.sample(
+                patience=self.patience, user_priors=False, ignore_fidelity=False
+            ) for _ in range(_n)]
+            for i, config in enumerate(new_configs):
+                config["id"].value = _new_configs[i]
+                config.fidelity.value = self.pipeline_space.fidelity.lower
```
karibbov, line 116: Combining this loop with the above list comprehension should add some speed up when `SAMPLES_TO_DRAW` is large

src/neps/api.py
```diff
@@ -198,9 +200,134 @@ def run(
...
+            logger=logger,
+            searcher=searcher,
+            searcher_path=searcher_path,
+            **searcher_kwargs,
+        )
+
+    # Used to create the yaml holding information about the searcher.
+    # Also important for testing and debugging the api.
+    searcher_info = {
+        "searcher_name": searcher,
```
TarekAbouChakra, line 241: Serializing 'searcher_info' which includes the 'searcher_name' as an instance, will lead to an error during YAML serialization in metahyper for ".optimizer_info.yaml". This error typically takes the form of a TypeError, signifying that the object is not JSON-serializable.
 
To resolve this issue, a quick workaround is to change the 'searcher_name' to a string during initialization:
```python:
    searcher_info = {
        "searcher_name": str(searcher), 
        "searcher_alg": searcher_alg,
        "user_defined_searcher": user_defined_searcher, 
        "searcher_args_user_modified": False,   
}
```
Keep in mind that this approach will result in 'searcher_name' being serialized as (assuming BayesianOptimization):
<neps.optimizers.bayesian_optimization.optimizer.BayesianOptimization object at 0xHEX...>
It's essential to consider alternative solutions for improving the serialization process.

src/neps/optimizers/bayesian_optimization/acquisition_samplers/freeze_thaw_sampler.py
```diff
@@ -67,31 +91,63 @@ def _sample_new(
...
+        if self.is_tabular:
+            _n = n if n is not None else self.SAMPLES_TO_DRAW
+            _partial_ids = {conf["id"].value for conf in partial_configs}
+            _all_ids = set(self.pipeline_space.custom_grid_table.index.values)
+            # accounting for unseen configs only
+            _n = min(_n, len(_all_ids - _partial_ids))
+            _new_configs = np.random.choice(
+                list(_all_ids - _partial_ids), size=_n, replace=False
+            )
+            new_configs = [__sample_single_new_tabular(i) for i in range(_n)]
```
herilalaina, line 127: Is there a way to get the entire tabular grid once? For me, this line can return duplicates in the sampled configurations.

karibbov: As far as my understanding goes, this should never return duplicates. Do you know if it returns multiple identical configurations in the `new_configs` or duplicate configs with already explored ones?

herilalaina: I was wrong, sorry. It doesn't return duplicates indeed.

src/neps/optimizers/bayesian_optimization/acquisition_functions/mf_ei.py
Neeratyoy: says `outdated` now and the code seems to have changed...

src/neps/optimizers/bayesian_optimization/acquisition_samplers/freeze_thaw_sampler.py
Neeratyoy: I think this is done, right
thanks!

Neeratyoy:
We can merge this if we have verified things and are satisfied.

PR #33, opened 2023-11-16
API refactoring + API tests and checks

TarekAbouChakra:
This PR focuses on NePS API refactoring and introduces several enhancements and tests:

- YAML generated to save optimizer information is now better configured, with specific details on the optimizer's arguments and the type of search done.
- Added test cases for each `neps.run` call type, covering Baseoptimizer, default YAML, and user YAML scenarios.
- Introduced a Metahyper test for the optimizer, evaluating multi-thread processing and pausing/resuming run scenarios. This test ensures that users won't unintentionally use different searchers for the same run, providing a more stable and predictable optimization environment.
- Added API tests on GitHub Actions.
- Accepts `None` to be passed for the pipeline space when the searcher is of type `BaseOptimizer`. Raises an error otherwise


src/neps/api.py
```diff
@@ -106,7 +108,7 @@ def run(
     ignore_errors: bool = False,
     loss_value_on_error: None | float = None,
     cost_value_on_error: None | float = None,
-    pre_load_hooks: List=[],
+    pre_load_hooks: List = [],
```
eddiebergman, line 111: Some information about typing, don't need to do anything here but thought it mgiht be good to share:

1. If you have `from __future__ import annotations` (which there is), you can use the lowercase `list`
2. [Do not use mutable arguments as a default](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).
3. We can type it better by following down where they're used, from what I gather it accepts `Callable[[metahyper.Sampler], metahyper.Sampler]`, where it can modify the sampler. Consider the typing can happen at a `neps` level and we only care about `class BaseOptimizer(metahyper.Sampler): ...`, then the full signature might look like:

```python
def run(
    ...,
    pre_load_hooks: list[Callable[[BaseOptimizer], BaseOptimizer]] | None = None
):
    if pre_load_hooks is None:
        pre_load_hooks = []
```

4. We can actually improve it even further using `typing.Iterable` to mean anything that we can iterate

```python
from typing import Iterable

def run(
    ...,
    pre_load_hooks: Iterable[Callable[[BaseOptimizer], BaseOptimizer]] | None = None
):
    if pre_load_hooks is None:
        pre_load_hooks = []
```

This means someone could pass a `tuple`, `set`, `generator` or really anything where we can go `for x in my_iterable: ...`.

5. Last bit is definitely over the top and low priority, but just to share some new typing tips. The function is actually a bit more specific in that it should return the **same** type of `BaseOptimizer` that went into it, not just any base optimizer. We can use `TypeVar`'s for this, which are essentially C+++ Templates or Java Generics.

```python
from typing import TypeVar

# This is some type that must be subclass of BaseOptimizer
OptimizerT = TypeVar("OptimizerT", bound=BaseOptimizer)

def run(
    ...,
    pre_load_hooks: Iterable[Callable[[OptimizerT], OptimizerT]] | None = None
):
    if pre_load_hooks is None:
        pre_load_hooks = []
```

@Neeratyoy thought you might also like to know about these things if you are reviewing

src/neps/api.py
```diff
@@ -200,24 +202,30 @@ def run(
...
+        "searcher_selection_source": "",
+        "searcher_modified_arguments": {},
+    }
+
     if isinstance(searcher, BaseOptimizer):
         searcher_instance = searcher
-        searcher_name = "custom"
-        searcher_alg = searcher.whoami()
-        user_defined_searcher = True
+        searcher_info["searcher_name"] = "custom"
```
eddiebergman, line 220: Won't this trigger for every kind of `BaseOptimizer`, even one's where we know their names?

src/neps/api.py
```diff
@@ -354,42 +334,46 @@ def _run_args(
...
+    searcher_info["searcher_name"] = searcher
+    searcher_info["searcher_alg"] = searcher_alg
 
     # Updating searcher arguments from searcher_kwargs
     for key, value in searcher_kwargs.items():
-        if user_defined_searcher:
+        if (
+            searcher_info["searcher_selection_source"] == "Default_Searcher-User_Choice"
+            or searcher_info["searcher_selection_source"] == "Custom-User_Yaml"
+        ):
```
eddiebergman, line 364: These use of string matching will lead to a bug at somepoint, I would recommend some constants.

src/neps/utils/common.py
```diff
@@ -231,6 +232,17 @@ def get_searcher_data(searcher: str, searcher_path: Path | str | None = None) ->
...
 
+def get_value(obj: Any):
+    if isinstance(obj, (str, int, float, bool, type(None))):
+        return obj
+    elif isinstance(obj, dict):
+        return {key: get_value(value) for key, value in obj.items()}
+    elif isinstance(obj, list):
+        return [get_value(item) for item in obj]
+    else:
+        return obj.__name__
```
eddiebergman, line 243: What's the point of this function in the larger context? I saw it above just to do printing. Seems to just convert most object values to their name, otherwise their raw value if `str(), int(), float(), bool() or None`?

If the goal is just to have strings or data that is serializable, might be better to rename it as `get_value` in `neps.util.commons` doesn't really speak to its use and will likely not get re-used by anyone else, as they won't know why or where to use it.

tests/test_neps_api/test_api.py
```diff
@@ -0,0 +1,204 @@
...
+import pytest
+import yaml
+
+
+# To change the working directly into the tmp_path when testing function
+@pytest.fixture(autouse=True)
+def use_tmpdir(tmp_path, request):
+    os.chdir(tmp_path)
+    yield
+    os.chdir(request.config.invocation_dir)
```
eddiebergman, line 15: Interesting, didn't know about this kind of hack. I would advise it's better in general to request the built in `tmp_path: Path` fixture to functions that need a directory to work from, but I guess that's not easily possible for some tests.

tests/test_neps_api/test_api.py
```diff
@@ -0,0 +1,204 @@
...
+                "hierarchy_consider": [0, 1, 2, 3],
+                "d_graph_features": 0,
+                "vectorial_features": None,
+            },
+        },
+    },
+    "priorband_neps_decided": {
+        "searcher_name": "priorband",
+        "searcher_alg": "priorband",
+        "searcher_selection_source": "Default_Searcher-NePS_Decision_Tree",
```
eddiebergman, line 67: Seeing these here, seems like they should be `ClassVar`'s on the objects themselves?

eddiebergman (approved):
Just left some minor comments, don't need to act on them really. I do like the idea that you test runnable examples through `pytest`, I think going forward this will make it easier to do _integration tests_ and make test writing easier for people.

If you are happy with others using the examples, you could move them to the root, `./examples` directory of the repo and still test them there. That way people will see them, use them, and their tested.

src/neps/utils/common.py
TarekAbouChakra: Some arguments are passed as instances or classes in neps.run like this example [here](https://github.com/automl/neps/blob/master/neps_examples/experimental/hierarchical_architecture_hierarchical_GP.py#L110-L140). When serializing it to the yaml, an error would occur.
Likewise when changing it to strings, something like <neps.optimizers.bayesian_optimization.optimizer.BayesianOptimization object at 0xHEX...> would then be saved. So I opted for a cleaner look by only saving the class name instead. 

This can probably be better if we allow the users to pass strings instead of classes and then we map those strings internally.

src/neps/api.py
TarekAbouChakra: True. Searcher name is currently the yaml file name where the searcher arguments and the algorithm type are located (If a user create a new yaml named `bo_new.yaml`, `searcher_name` would then be `bo_new` and `searcher_algorithm` would be `BayesianOptimization`). For BaseOptimizer case, since the searcher is already passed as an instance, the adjustment to searcher_name being `custom` was done.

tests/test_neps_api/test_api.py
TarekAbouChakra: I took the inspiration from this test [here](https://github.com/automl/neps/blob/master/tests/test_examples.py#L9-L13).

tests/test_neps_api/test_api.py
TarekAbouChakra: I changed these to have solution_yamls, which is easier to then locate and add for future tests I assumed.

Neeratyoy (approved):

PR #36, opened 2023-11-17
typing: Export `py.typed` for `neps` + `metahyper`

eddiebergman:
This allows consumers who using `mypy` to type check their code against neps and metahyper.

See: https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages


eddiebergman:
Took that as a sign to merge ;)

PR #40, opened 2023-11-21
ci(dependancies): Fix tests by using `extras` and check for specific `MissingDependancyError`

eddiebergman:
Supersedes #39, please refer to there

This follows the poetry guide here for properly installing additional dependancies:
https://python-poetry.org/docs/master/pyproject/#extras

Also makes sure that these are installed in the git workflow as these run the examples.

Extra comments:
---

Testing things locally, you would not know this is an issue, except when the CI runs. There are two options for this. Everything is done in a PR so it can be seen before merging, or all tests are enabled locally and contributors are expected to run and fix them.


src/metahyper/utils.py
```diff
@@ -115,6 +117,10 @@ def instance_from_map(
         ValueError: if the request is invalid (not a string if allow_any is False),
             or invalid key.
     """
+    if isinstance(request, MissingDependencyError):
+        # This happens when some optional dependancy is missing, the error
+        # message will signal to the user what to do
+        raise request
```
karibbov, line 123: Shouldn't we check for the type of the `mapping[request]` here (or `instance` later on), instead of the `request`, which is a key from the `mapping`? Since we have the mapping: `SurrogateModelMapping = {"deep_gp": DeepGP, ...}`

tests/test_metahyper/test_instance_mapping.py
```diff
@@ -0,0 +1,24 @@
...
+        ),
+        MissingDependencyError(
+            libname="neps", dep="test_dep_2", install_group=None, cause=ImportError()
+        ),
+    ],
+)
+@pytest.mark.metahyper
+def test_missing_dependancy_gets_flagged(err: MissingDependencyError) -> None:
+    with pytest.raises(MissingDependencyError, match=err.dep):
+        instance_from_map(mapping={}, request=err)
```
karibbov, line 24: [This](https://github.com/automl/neps/blob/f6cecbf6ae410fa45dc9a5e71424e88b3adfc4f0/src/neps/optimizers/bayesian_optimization/optimizer.py#L156-L161) is how we usually call `instance_from_map(...)` function to get the surrogate model, the `surrogate_model` parameter is usually a string

src/metahyper/utils.py
eddiebergman: Oh you're absolutely right :) Will fix

tests/test_metahyper/test_instance_mapping.py
eddiebergman: Thanks for the reference, super useful!

eddiebergman:
Closingg, can re-address as necessary

PR #46, opened 2023-12-04
Pipeline_space as yaml input

Neeratyoy:
Enabled initialization of search_space via a YAML file. 
You can find an example in neps_examples/basic_usage/defining_search_space. For more information, refer to docs/pipeline_space.md.
Supported Parameter Types: Int, Float, Categorical, Constant


Neeratyoy:
The directories were restructured. Needs to updated with the `master`.

neps/api.py
```diff
@@ -311,6 +317,9 @@ def _run_args(
         # Support pipeline space as ConfigurationSpace definition
         if isinstance(pipeline_space, CS.ConfigurationSpace):
             pipeline_space = pipeline_space_from_configspace(pipeline_space)
+        # Support pipeline space as YAML file
+        elif isinstance(pipeline_space, str):
+            pipeline_space = pipeline_space_from_yaml(pipeline_space)
```
TarekAbouChakra, line 322: Although `Path` is also supported as a type to be passed in the `pipeline_space` argument. There is no condition for it here and how it should be handled internally.

Currently having a `Path` type pipeline_space could result in an error because it will skip these checks.

neps/search_spaces/yaml_search_space_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        super().__init__(self.message)
+
+
+def deduce_and_validate_param_type(name, details):
+    """
+    Deduces the parameter type from details and validates them.
+
+    Args:
+        name (str): The name of the parameter.
+        details (dict): A dictionary containing parameter specifications.
```
TarekAbouChakra, line 63: Might be a bit nitpicking here but since the argument types are available in the comments would it also make sense to have them in the function arguments themselves ?
ex:
```python
def deduce_and_validate_param_type(name: str, details: dict):
```
Or: 
```python
from typing import Dict

def deduce_and_validate_param_type(name: str, details: Dict[str, int]): # Any types required
```

tests/test_yaml_search_space/correct_config.yaml
```diff
@@ -0,0 +1,31 @@
...
+    lower: 0.00001
+    upper: 0.1
+    log: TRUE
+    is_fidelity: off
+
+  num_epochs:
+    lower: -3
+    upper: 30
+    log: false
+    is_fidelity: on
```
TarekAbouChakra, line 12: This looks a bit weird, could we have another parameter name for this ?

danrgll: parameter naming is now fixed [here](https://github.com/automl/neps/pull/46/commits/15d4cc885e44f673eaaa6fd2f1b1e16ec6ffaae2)

neps/api.py
danrgll: The usage of the Path object from the pathlib module is now supported [here](https://github.com/automl/neps/pull/46/commits/c70b85ee5c65ed178b84e7c11e94ed8110e2b6ab)

neps/search_spaces/search_space.py
```diff
@@ -61,6 +66,116 @@ def pipeline_space_from_configspace(
...
+    """
+    Reads configuration details from a YAML file and constructs a pipeline space
+    dictionary.
+
+    This function extracts parameter configurations from a YAML file, validating and
+    translating them into corresponding parameter objects. The resulting dictionary
+    maps parameter names to their respective configuration objects.
+
+    Args:
+        yaml_file_path (str): Path to the YAML file containing parameter configurations.
```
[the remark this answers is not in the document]
danrgll: The usage of the Path object from the pathlib module is now supported. I added it also to the Docstring[here](https://github.com/automl/neps/pull/46/commits/c70b85ee5c65ed178b84e7c11e94ed8110e2b6ab)

neps/search_spaces/yaml_search_space_utils.py
danrgll: Agree, added it + some missing details in some parameter validation functions[here](https://github.com/automl/neps/pull/46/commits/91d8a457db76cee1970fbabd6941325c656b758e)

TarekAbouChakra (approved):
The feature addition looks good. I've tested the example, and it worked as expected. 

The inclusion of testing is a great addition, adding robustness to the codebase.

danrgll:
Found two issues:

- Documentation gets not formatted correctly for mkdocs. 
- And new argument types for pipeline Space gets not detected by (PyCharm)

I was able to fix both problems, but during commit the files get reformatted and results back in the same issue. Does committing with --no-verify flag overcome this automated reformatting?

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -0,0 +1,159 @@
...
+        "tests/test_yaml_search_space/correct_config_including_priors.yml"
+    )
+    assert isinstance(pipeline_space, dict)
+    assert isinstance(pipeline_space["learning_rate"], FloatParameter)
+    assert pipeline_space["learning_rate"].lower == 0.00001
+    assert pipeline_space["learning_rate"].upper == 0.1
+    assert pipeline_space["learning_rate"].log is True
+    assert pipeline_space["learning_rate"].is_fidelity is False
+    assert pipeline_space["learning_rate"].default == 3.3e-2
+    assert pipeline_space["learning_rate"].default_confidence_score == 0.125
```
Neeratyoy, line 77: @danrgll @TarekAbouChakra 
For this PR let us push this. 

For later refactoring, can we compress such lines by creating a `FloatParameter` using these values explicitly in one line, and comparing the loaded instantiation using something [like this](https://github.com/automl/neps/blob/master/neps/search_spaces/hyperparameters/float.py#L48)?
If so, let us keep a note and do this for next year's release.

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -0,0 +1,159 @@
...
+    test_correct_yaml_file(
+        "tests/test_yaml_search_space/correct_config_including_types" ".yaml"
+    )
+
+
+@pytest.mark.neps_api
+def test_correct_including_priors_yaml_file():
+    """Test the function with a correctly formatted YAML file."""
+    pipeline_space = pipeline_space_from_yaml(
+        "tests/test_yaml_search_space/correct_config_including_priors.yml"
```
Neeratyoy, line 68: Could we perhaps make the location (`tests/test_yaml_search_space/`) a global or class variable for all such relative path imports?
(Same, please note for the next release)

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -0,0 +1,159 @@
...
+    with pytest.raises(SearchSpaceFromYamlFileError) as excinfo:
+        pipeline_space_from_yaml(
+            "tests/test_yaml_search_space/inconsistent_types_config.yml"
+        )
+    assert str(excinfo.value.exception_type == "TypeError")
+    with pytest.raises(SearchSpaceFromYamlFileError) as excinfo:
+        pipeline_space_from_yaml(
+            Path("tests/test_yaml_search_space/inconsistent_types_config2.yml")
+        )
+    assert str(excinfo.value.exception_type == "TypeError")
```
Neeratyoy, line 126: Since a boolean expression is converted to either "True" or "False" the `assertion` will always be True.
Are we sure these test cases pass? 
We should check these before releasing them?

Neeratyoy:
Looks good!

There is only one issue with the tests cases for the wrong YAML inputs where I think the test is not being checked correctly.

Overall, it looks quite comprehensive with what looks like adequate documentation! 
Some other comments I have left as notes for future improvements.

One more such point is, can we now generate a yaml from a `pipeline_space` or `SearchSpace` and save it along with the files we write for a NePS run?
If so, let us add that too @danrgll @TarekAbouChakra.

tests/test_yaml_search_space/test_search_space.py
danrgll: You are right. im going to fix that.

danrgll: Should be fixed[here](https://github.com/automl/neps/pull/46/commits/075c79b67944a881b6153d2f1e79ac4633ef73ae)

Neeratyoy:
@TarekAbouChakra @danrgll 
Please feel free to coordinate and monitor merges during December. :)

PR #53, opened 2024-01-27
fix(metahyper): Print useful information in logs for exception occuring

eddiebergman:
This is more of a bandaid fix such that the error/traceback shows up at least somewhere. For debugging pipelines and initial setup, there should be some option where the error just bubbles all the way up and crashes the program.

### DIFF
--- a/neps/metahyper/api.py
+++ b/neps/metahyper/api.py
@@ -4,6 +4,7 @@ import inspect
 import logging
 import shutil
 import time
+import traceback
 import warnings
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


neps/metahyper/api.py
TarekAbouChakra, line 1: This is indeed an essential feature. However, I'm a bit puzzled, shouldn't the traceback already be logged with the outer [try-except](https://github.com/automl/neps/pull/53/files#diff-91d12133e2e563246e4eef974b69c6bce380fb17c651119b8bb7c5358ee064c6R413-R417) statement, especially since we're using [`logger.exception`](https://stackoverflow.com/questions/5191830/how-do-i-log-a-python-error-with-debug-information)? If I'm mistaken or if this behaves differently in certain scenarios, please correct me.

Additionally, I'm wondering if it might be more appropriate to catch the error from the ['except statement'](https://github.com/automl/neps/pull/53/files#diff-91d12133e2e563246e4eef974b69c6bce380fb17c651119b8bb7c5358ee064c6R413-R417) of the initial try block as well. What are your thoughts on this?

eddiebergman:
I learned something new today, I did not know `logger.exception` does that by default! Good to know :)

In that case, I think the main feature that should be added is to give an option to raise immediatly on failure. This makes developing easier, i.e. no one will get their evaluation function right first time.

As an addendum, I also think that the error logic in the parsing of the result should always mention what it got as the result that it failed to parse. Simply saying something didn't work isn't very informative

I will close this PR as that seems like two things:
* A feature for `on_error: Literal["continue", "raise"]` or something like that to help initial development. Maybe even make it default "raise" as that's what is most likely needed while initially trying things
* Better error messages saying what it got as the result such that parsing it out failed.

Issue #42, opened 2023-11-22
Metahyper sampling extra config


TarekAbouChakra:
Metahyper is sometimes sampling an extra configuration, causing tests to fail at some occasions:

https://github.com/automl/neps/actions/runs/6894554381/job/18756615811
https://github.com/automl/neps/actions/runs/6213981866/job/16865511613
https://github.com/automl/neps/actions/runs/6059364933/job/16442416911

Difficult to reproduce, cause of problem is still known, example that the problem  appears in: https://github.com/automl/neps/blob/master/neps_examples/basic_usage/hyperparameters.py

eddiebergman:
Pretty sure this was solved in issue #71, closing for now

PR #54, opened 2024-02-02
Refactored Metahyper and Its Test Cases

TarekAbouChakra:
The primary objective of this pull request is to address the issue related to Metahyper sampling an additional configuration. 
- #42.

Changes Made:
- Modified the Metahyper code to ensure it is in a locked state during execution in multi-threading (previously the _check_max_evaluations function had code that was not in a locked state) to prevent potential race conditions.
- Addressed instances where the internal read function in Metahyper was called twice by optimizing it to read only once to fulfill all purposes.
- Cleared out ConfigInRun data after the completion of configuration evaluation and post hook execution.
- Added sleep times for all locking operations when the lock is unavailable for stability.

Changes in Tests:
- Adopted concurrent.futures as the multi-threading library in the tests, providing better control over processes for improved debugging and scalability when needed.

### DIFF
--- a/neps/metahyper/api.py
+++ b/neps/metahyper/api.py
@@ -49,6 +49,13 @@ class ConfigInRun:
         ConfigInRun.previous_pipeline_directory = previous_pipeline_directory
         ConfigInRun.optimization_dir = Path(optimization_dir)
 
+    @staticmethod
+    def empty_config_data():
+        ConfigInRun.config = None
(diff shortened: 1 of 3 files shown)


eddiebergman:
Heyo @TarekAbouChakra, I just merged #71 which reworked metagyper and converted it to `neps.runtime`. I'm going to look through these changes and turn it into a new clean PR!

PR #55, opened 2024-02-04
Refactored yaml pipeline_space

danrgll:
These proposals were implemented:
- add global variable to handle path location in yaml tests
- Parameter comparison in yaml tests through updated __eq__ function resulting in desired one line check

Added Boolean check in parameter classes float/categorical for log and is_fidelity to avoid interpret typing error wrong in yaml file + tests

Added logging info to inform the user that his e notation gets converted to float(parameter)

Added constraints for default value (have to be in range for float/int parameter) or in choices for categorical parameter + tests

Fixed Link in Docs


neps/search_spaces/hyperparameters/categorical.py
```diff
@@ -39,14 +39,39 @@ def __init__(
...
+            raise ValueError(
+                f"Default value {default} is not in the provided choices {self.choices}"
+            )
+
+        # Check if 'is_fidelity' is a boolean
+        if not isinstance(is_fidelity, bool):
+            raise TypeError(
+                f"Expected 'is_fidelity' to be a boolean, but got type: "
+                f"{type(is_fidelity).__name__}"
+            )
```
eddiebergman, line 59: Nothing to change, my general opinion is that there's types for a reason and you can't check every input that it's of the correct type. If conflicted on whether to check it or not, I usually favor on not.

However something like `if not 0 <= value <= 1` should be checked as it can't be conveyed in a type.

neps/search_spaces/search_space.py
```diff
@@ -177,7 +183,7 @@ def pipeline_space_from_yaml(
                     "For categorical parameter: cat, categorical\n"
                     "For constant parameter: const, constant\n"
                 )
-    except (KeyError, TypeError, ValueError) as e:
+    except (KeyError, TypeError, ValueError, FileNotFoundError) as e:
         raise SearchSpaceFromYamlFileError(e) from e
     return pipeline_space
```
eddiebergman, line 189: This whole block can really just be a series of `try: excepts` rather than a nest `try: except`. No need to change now, just for your info

neps/search_spaces/yaml_search_space_utils.py
```diff
@@ -1,10 +1,13 @@
...
 
+import logging
 import re
 
+logger = logging.getLogger(__name__)
+
 
 def convert_scientific_notation(value: str | int | float, show_usage_flag=False) \
-                                -> float | (float, bool):
+        -> float | (float, bool):
```
eddiebergman, line 10: First, use `tuple[float, bool]`, doing `(float, bool)` isn't valid syntax.

Also try to avoid `\` in code. Normally formatters and linters will fix this up in the following way which I believe is more readable for longer function signatures.

```python
def convert_scientific_notation(
    value: str | int | float,
    show_usage_flag: bool = False
) -> float | tuple[float, bool]:

File: tests/test_yaml_search_space/test_search_space.py [hunk 4]
```diff
@@ -8,105 +8,64 @@
...
+        assert float2.__eq__(pipeline_space["param_float2"]) is True
+        cat1 = CategoricalParameter([2, "sgd", 10e-3], False)
+        assert cat1.__eq__(pipeline_space["param_cat"]) is True
+        const1 = ConstantParameter(0.5, False)
+        assert const1.__eq__(pipeline_space["param_const1"]) is True
+        const2 = ConstantParameter(1e3, True)
+        assert const2.__eq__(pipeline_space["param_const2"]) is True
+
+    test_correct_yaml_file(BASE_PATH + "correct_config.yaml")
+    test_correct_yaml_file(BASE_PATH + "correct_config_including_types.yaml")
```
eddiebergman, line 36: Three things on code quality improvements:
1. `assert something() is True` is equivalent to `assert something()`
2. `a.__eq__(b)` is equivalent to `a == b`
3. You can use `pytest.mark.parametrize(...)` to essentially run this test twice, each with a different yaml file.

I would personally write this in the following way:
```python
@pytest.mark.neps_api
@pytest.mark.parametrize(
	"path",
	[
		BASE_PATH + "correct_config.yaml",
		BASE_PATH + "correct_config_including_types.yaml"
	]
)
def test_correct_yaml_files(path: str):
	"""Test the function with a correctly formatted YAML file."""
	pipeline_space = pipeline_space_from_yaml(path)
	assert isinstance(pipeline_space, dict)
	
	float1 = FloatParameter(0.00001, 0.1, True, False)
	int1 = IntegerParameter(-3, 30, False, True)
	int2 = IntegerParameter(100, 30000, True, False)
	cat1 = CategoricalParameter([2, "sgd", 10e-3], False)
	float2 = FloatParameter(3.3e-5, 0.15, False, False)
	const1 = ConstantParameter(0.5, False)
	const2 = ConstantParameter(1e3, True)

	assert pipeline_space["param_float1"] == float1
	assert pipeline_space["param_int1"] == int1
	assert pipeline_space["param_int2"] == int2	
	assert pipeline_space["param_float2"]) == float2
	assert pipeline_space["param_cat"] == cat1
	assert pipeline_space["param_const1"] == const1
	assert pipeline_space["param_const2"] == const2
```

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -8,105 +8,64 @@
...
-    assert pipeline_space["dropout_rate"].value == 1e3
-    assert pipeline_space["dropout_rate"].default == 1e3
+    float1 = FloatParameter(0.00001, 0.1, True, False, 3.3e-2, "high")
+    assert float1.__eq__(pipeline_space["learning_rate"]) is True
+    int1 = IntegerParameter(3, 30, False, True, 10)
+    assert int1.__eq__(pipeline_space["num_epochs"]) is True
+    cat1 = CategoricalParameter(["adam", 90e-3, "rmsprop"], False, 90e-3, "medium")
+    assert cat1.__eq__(pipeline_space["optimizer"]) is True
+    const1 = ConstantParameter(1e3, True)
+    assert const1.__eq__(pipeline_space["dropout_rate"]) is True
```
eddiebergman, line 53: Refer to the previous suggestion, however you don't need the parametrize here

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -8,105 +8,64 @@
...
 
 @pytest.mark.neps_api
 def test_incorrect_yaml_file():
     """Test the function with an incorrectly formatted YAML file."""
     with pytest.raises(SearchSpaceFromYamlFileError) as excinfo:
-        pipeline_space_from_yaml(
-            Path("tests/test_yaml_search_space/incorrect_config.txt")
-        )
+        pipeline_space_from_yaml(Path(BASE_PATH + "incorrect_config.txt"))
     assert excinfo.value.exception_type == "ValueError"
```
eddiebergman, line 61: I feel like this second `assert excinfo.value.exception_type == "ValueError"` isn't needed. You've nicely given a specific error type of `SearchSpaceFromYamlFileError`. Check that this error type is also a `ValueError` is not what the test is about.

tests/test_yaml_search_space/test_search_space.py
```diff
@@ -142,18 +95,62 @@ def test_yaml_file_including_unkown_types():
...
+def test_float_is_fidelity_not_boolean():
+    """Test if an exception is raised when for FloatParameter the 'is_fidelity'
+    attribute is not a boolean."""
     with pytest.raises(SearchSpaceFromYamlFileError) as excinfo:
         pipeline_space_from_yaml(
-            "tests/test_yaml_search_space/not_allowed_key_config.yml"
+            BASE_PATH + "not_boolean_type_is_fidelity_float_config.yaml"
         )
-    assert excinfo.value.exception_type == "KeyError"
+    assert excinfo.value.exception_type == "TypeError"
```
eddiebergman, line 136: Please see above comment about the additional assert

eddiebergman (changes requested):

PR #96, opened 2024-05-16
Integrate pipeline space into run args yaml + new design for defining constant parameter

danrgll:
- Declarative usage doc changes
- Define pipeline space in run args yaml
- New design for constant parameter (const1: value)
- Adapt search_space_from_yaml code to new functionalities


neps/search_spaces/search_space.py
```diff
@@ -90,107 +94,80 @@ def pipeline_space_from_configspace(
...
+def pipeline_space_from_yaml(  # noqa: C901, PLR0912
+    config: str | Path | dict,
+) -> dict[str, Parameter]:
+    """Reads configuration details from a YAML file or a dictionary and constructs a
+    pipeline space dictionary.
 
     Args:
-        yaml_file_path (str | Path): Path to the YAML file containing parameter
-        configurations.
+        config (str | Path | dict): Path to the YAML file or a dictionary containing
```
eddiebergman, line 104: Just fyi, you don't need to specify the types in docstrings anymore, it just pulls it straight from the function definition!

neps/search_spaces/search_space.py
```diff
@@ -90,107 +94,80 @@ def pipeline_space_from_configspace(
...
-                pipeline_space[name] = IntegerParameter(
-                    lower=details["lower"],
-                    upper=details["upper"],
-                    log=details.get("log", False),
-                    is_fidelity=details.get("is_fidelity", False),
-                    default=details.get("default", None),
-                    default_confidence=details.get("default_confidence", "low"),
-                )
+                formatted_details = formatting_int(name, details)
+                pipeline_space[name] = IntegerParameter(**formatted_details)
```
eddiebergman, line 149: Nice, much more concise :)

neps/search_spaces/yaml_search_space_utils.py
```diff
@@ -82,32 +81,44 @@ def __init__(self, exception):
...
+                param_type = deduce_param_type(name, details)
+    except TypeError as e:
+        # because details could be int, float
+        if isinstance(details, (str, int, float)):
+            param_type = "const"
+            return param_type
+        else:
+            raise TypeError(
+                f"Unable to deduce parameter type for '{name}' with details '{details}'.") \
+                from e
```
eddiebergman, line 121: I think it would be much cleaner and easier to understand this by starting with something like:

```python
if isinstance(details, (str, int, float)):
   ... # assume constant
   return something
   
if isinstance(details, dict):
   ... # It's a dict, do what you do
   return something
   
raise TypeError() # Got something we didn't expect to get
```

A lot less try/catches and it's clearer there's a seperate path for each possibility.

```

File: neps/utils/run_args_from_yaml.py [hunk 4]
```diff
@@ -256,6 +256,34 @@ def process_config_key(settings: Dict, special_configs: Dict, keys: List) -> Non
                     )
 
 
+def process_pipeline_space(key, special_configs, settings):
+    if special_configs.get(key) is not None:
+        pipeline_space = special_configs[key]
```
eddiebergman, line 261: Never seen this combination before funnily enough, here are someother for you info:

```python
if key in special_configs:
    pipeline_space = special_configs[key]
    
## OR

pipeline_space = special_configs.get(key)
if pipeline_space is not None:
    ... # Use pipeline_space as you would
    
# OR

if (pipeline_space := special_configs.get(key)) is not None:
   # Use pipeline space as you would
```

This function is likely going to run only once and performance doesn't matter, choose your favorite :)

eddiebergman (approved):
Looks good, I left some minor commentary which I'm not going to block anything on. Seems things are a lot more concise and much easier to read, nice one!

The one thing I haven't done is render the docs yet (bad internet), I will do so tomorrow and scroll through it. If all is good, I will merge!

neps/search_spaces/yaml_search_space_utils.py
danrgll: The idea was in the first place to be more efficient because of the less usage of constant parameter but then I got lost in all these cases. I changed it in the new PR

neps/utils/run_args_from_yaml.py
danrgll: yeah its like this: 

```python
pipeline_space = special_configs.get(key)
if pipeline_space is not None:
```
The key should be in the dict by default with the value None thats why the first solution would not work. I could simplify it to:

```python
if special_configs[key] is not None:
pipeline_space = special_configs[key]
```

danrgll: with .get I get for sure a None as a return, if special_configs not contains this key

PR #107, opened 2024-06-18
Update pyproject.toml for PyTorch versions

Neeratyoy:


eddiebergman:
Good enough given what we've got I would say, if it causes issues. Then these issues should have an associated test so we can catch them in the future.

Neeratyoy:
@eddiebergman 
Do you wanna merge or I go with it?

karibbov:
Updating `torch > 2.0.0` requires `torchvision >= 0.16`. Then, we get error from `import torchvision`. Anyone can decrypt the traceback?

karibbov:
Turns out [AttributeError: 'int' object has no attribute 'nodes' ](https://github.com/automl/neps/actions/runs/9449148188/job/26024747489?pr=106)
originates from [here](https://github.com/automl/neps/blob/0553bd54dda617286eaf77599c77678b2a94cc75/neps/optimizers/bayesian_optimization/kernels/utils.py#L38-L39). When the dimension of the nested list -
`graphs` is 2  `_nested_graphs.ndims` sometimes returns `3` if the graphs in the dimension 2 are consistent, which causes the if clause to be taken and graphs list to be corrupted. 
Made a short fix to it [here](https://github.com/automl/neps/pull/107/commits/ade69db7c23673fcbe9d2ac2b6dc348810b45bd0), doesn't seem to break any tests

Neeratyoy:
That's a great catch!

karibbov (approved):

eddiebergman (approved):

Issue #74, opened 2024-04-25
Develop CLI for simplified yaml file management + running neps


danrgll:
Ideally, follows after #65. 

Possible usage:

- `neps help`: Displays available arguments and their descriptions to guide users.
- `neps init`: E.g. automatically creates a 'run_args' YAML configuration file, pre-filled with keys (defaults provided where applicable) and leaves required fields empty for user input. This simplifies the initial setup by eliminating the need for users to manually create the YAML structure.
- `neps run neps_config.yaml`: Enables executing neps.run() directly via a command line, streamlining the process of running configurations.

PR #117, opened 2024-07-30
Introduce CLI

danrgll:
Introduce the first commands of the CLI.
- neps run "path/to/run_args.yaml" --root_directory "path/to/root_directory" --overwrite-working-directory ....
- neps init --config-path "path/to/config.yaml"


danrgll:
Closes #74

PR #134, opened 2024-08-13
refactor: BO and ifBO

eddiebergman:
This PR simplifies and speedsup/improves BO and ifBO. By no means is this list of things exhaustive but it covers some of the major changes and new toys present.

How?

---

This was primarily done by only using the `SearchSpace` for its definitions, not it's methods. Interacting with models that expect tensors, we focus on encoding directly to a tensor, and acting directly on this encoded space, instead of going back and forth between `SearchSpace` and the data format that the surrogate models expect.

**Before** pass around `list[SearchSpace]` and each component encodes as needed, often performing operations directly on the `SearchSpace`.

**After**: Encode the `list[SearchSpace]` into what's required and inform the components about the encoding.

This buys a lot of time to perform better acquisition optimization, as well as avoids bloating the ever growing list of methods in `SearchSpace`, which can not provide a solution for every kind of model/optimizer we have.

---

As part of this, we now use `botorch` as a dependancy, which is primarily built on top of `gpytorch`, which we already depended on. Some of the benefits include:

* No hand rolled kernels (potentially buggy, less code to maintain)
* Fully differentiable gp/kernel operations as well as the option to set GP hyperparameter priors. I hand-coded some priors based on prior work, such as [Carl Hvafners work](https://arxiv.org/abs/2402.02229), but recently this prior actually became the defaults for botorch and means less for us to maintain. We now do MAP optimization of GP hyperparameters instead of MLE optimization which was vastly more stable and consistent in some toy experiments. Their optimization for it out-performed anything I could hand-roll prior.
* We now have access to the full suite of botorch acquisitions, some including numerical stability tricks, taking care of Pending configurations without fantazisation, batching, batch acquisition and a lot more. One interesting thing to note is that this would make trasitioning model based methods to multi-objective much more trivial as they directly support MO in their models and acquisition.
* One area where I do handroll some things is a [`class WeightedAcquisition`](https://github.com/automl/neps/blob/refactor-acq-refresh/neps/optimizers/bayesian_optimization/acquisition_functions/weighted_acquisition.py), which can take in a botorch `AcquisitionFunction` and apply a custom weighting to the output. For example, [here](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/acquisition_functions/pibo.py#L33-L64) is a function that wraps an arbitrary acquisition function, and applies a weight based on the pdf of the samples under a prior, i.e. PiBo. You can see it in use [here](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/models/gp.py#L351-L362)
* This made acquistion optimization much easier. We now scale from roughly a measily 50ish samples, to ~1000 samples, including some rounds of gradient based optimization, restarts and more.

---

Also, I have removed a lot of the _fake_ flexibility that was offered for BO and ifBO. The primary thing removed is that our hand-rolled `GP` and the ftpfn model, are not attempted to be treated the same. They share very little in common, are acquired from in very different manners and have very different data-encodings. With the removal of DeepGP, these are our only two surrogates and we just treat them as two very different things. Maybe we try to unify in the future but I do not know what we gain from that.

In reality, we as developers would be the only one to use more advanced options and in general, they would be confusing to users actually looking to configure them, let alone the fact passing custom objects or even some of our own classes/objects would not work. Maybe we introduce the flexibilty at some point but it obfuscated the code, made it harder to maintain, test and debug. As an example, both ifBO and BO now only have one method `ask()` which contains most of the logic you would expect to see when referencing a paper/description of the algorithm.

Here is the `ask()` of both [BO](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/optimizer.py#L137-L216) and [ifBO](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/multi_fidelity/ifbo.py#L163-L288) now, which removed most of the abstractions and is now just direct function calls. It also removed the two step `load_configs()` and `get_next_config()` that we had before.

The result of this is that using the models is now "stateless", and mostly accessible through a function call.

* [`make_default_single_obj_gp(...)`](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/models/gp.py#L69-L128)
* [`encode_trials_for_gp(...)`](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/models/gp.py#L221-L261)
* [`fit_and_acquire_from_gp(...)`](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/optimizers/bayesian_optimization/models/gp.py#L264)

ifBO is fairly similar in terms of the function calls.

---

As representing configurations as a complex object `SearchSpace` is highly innefficient for some of the model routines, such as encoding/decoding/sampling/acquisiton-function optimization, I avoid the use of the methods present in `SearchSpace`, and treat it as just a definition of hyperparameters. **Instead, we define an encoding and encode all configuration information into one big tensor.** The encoder can translate back and forth:

Conceptually, `list[SearchSpace] <-> list[dict] <-> Encoder <-> Tensor`

Doing so meant that we go from "asking the search space to sample itself and then do all transformations" to "Sample a tensor and do tensor operations to match the encoding". No objects, little python, just torch.

This required some new infrastructure that was aware of how configurations are encoded ([`ConfigEncoder`](https://github.com/automl/neps/blob/39ffe48a6db0ed53227356ea422721b5ed2d6843/neps/search_spaces/encoding.py#L214-L240)).

This most important piece of new infrastructure is the `Domain`.

* [`Domain`](https://github.com/automl/neps/blob/refactor-acq-refresh/neps/search_spaces/domain.py): A dataclass that represents a numeric range, dtype, whether it's binned, log scale etc... The most important method is `cast()`, which allows you to convert between domains, e.g. cast from `Domain.floating(10, 10_000, log=True)` to `Domain.floating(0, 1, bins=18)`
    
```python
domain1 = Domain.floating(10, 10_000, log=True)
x_in_domain_1 = torch.tensor([10, 100, 1_000, 10_000])

domain2 = Domain.floating(0, 1, bins=18)
x_in_domain_2 = domain2.cast(x, frm=domain1)
```
   
Anywhere where we use a tensor, there is a `Domain` associated with it somehow.
In summary, it contains information about what kind of numbers are in the tensor.
We have them in quite a few places and put to good use:

* `ConfigEncoder.domains: list[Domain]`, one domain for each column in the tensor representing encoded configs.
* `XXXParameter`, gives information about the domain of parameter outputs
* `TorchDistributionWithDomain`, as dumb as it sounds, it combines a torch distribution with the domain over which is has support/samples over.
* Samplers (below), samples some values in it's own domain and takes in a `Domain  | list[Domain]` into which you'd like those samples transformed.
* Priors (below), takes in a tensor and its domain, then can transform it into the space of its prior distribution and calculate the pdf.
 
---

* [`Sampler(Protocol)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/samplers.py#L24-L110): A Protocol for something that can sample tensors. Related, is also the protocol [`class Prior(Sampler)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/priors.py#L34), which extends a `Sampler` by being able to also calculate `log_probs` a tensor of configs, used in things like pibo acquisition and prior based sampling. The mains once currently there are:
    * [`Sobol(Sampler)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/samplers.py#L115)
    * [`Uniform(Prior)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/priors.py#L378)
    * [`CenteredPrior(Prior)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/priors.py#L269), which handle parameters with and without defaults jointly.
    * [`WeightedPrior(Prior)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/priors.py#L421), which allows you to combine multiple priors by weights.
    * [`WeightedSampler(Sampler)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/samplers.py#L173), which is the same but for samplers which are not prior enabled.
    * [`BorderSampler(Sampler)`](https://github.com/automl/neps/blob/07eb8f2c347f3bc645804fc45ebf9e2e2608f66a/neps/sampling/samplers.py#L273), which efficiently generates border configurations.
        
The primary method is pretty straight forward. The most important argument is really `to=`, which lets you say "in what domain(s) would you like your samples?" This means you can sample a big tensor of uniform and convert it directly into the domain of encoded configs, (i.e. integers for categoricals, min-max normalized floats/ints, etc...).
        
```python
def sample(
    self,
    n: int | torch.Size,
    *,
    to: Domain | list[Domain],
    seed: torch.Generator | None = None,
    device: torch.device | None = None,
    dtype: torch.dtype | None = None,
) -> torch.Tensor:
    """Sample `n` points and convert them to the given domain.

    Args:
        n: The number of points to sample. If a torch.Size, an additional dimension
            will be added with [`.ncols`][neps.samplers.Sampler.ncols].
            For example, if `n = 5`, the output will be `(5, ncols)`. If
            `n = (5, 3)`, the output will be `(5, 3, ncols)`.
        to: If a single domain, `.ncols` columns will be produced form that one
            domain. If a list of domains, then it must have the same length as the
            number of columns, with each column being in the corresponding domain.
        seed: The seed generator
        dtype: The dtype of the output tensor.
        device: The device to cast the samples to.

    Returns:
        A tensor of (n, ndim) points sampled cast to the given domain.
    """
    ...
```

Most of the `Prior`'s are back by torch distributions, which has the aptly named `TorchDistributionWithDomain`, which encapsulates both a distribution and the domain over which it samples. The `cast()` method allows fluidly transforming between distribution domains, sample domains, and config encoding domains.
        
---

For some future work, I believe many of the bandit prior methods could benefit from the `Prior` class, as it allows calculating priors over both uniform parameters and those with a prespecified `default`.


eddiebergman:
Lol tests pass, I guess non of our tests anywhere hit this because this definitely shouldn't work right now

karibbov:
There are some minor updates on the [mergeDyHPO](https://github.com/automl/neps/blob/d6e123faf41dde2461241a20b2014e4345d9a2a3/neps/optimizers/bayesian_optimization/models/deepGP.py) branch which I'll go over with some comments to follow. I don't think `NotPSDError` needs any special addressing, it usually happens to rise when the model is fed bad data (as in many repeating or similar data points, perhaps even many zero values) so it should occur fairly seldom.

neps/optimizers/models/deepGP.py
```diff
@@ -0,0 +1,572 @@
...
+
+        return lc_buffer
+
+    def encode_y(
+        self, y: list[float]
+    ) -> tuple[torch.Tensor, None | tuple[int | float, int | float]]:
+        tensor = torch.tensor(y, device=self.device, dtype=torch.float32)
+        if self.fidelity_bounds:
+            _min, _max = tensor.min(), tensor.max()
+            tensor.sub_(_min).div_(_max - _min)
```
karibbov, line 263: I think we should drop output normalization, I've added it as an option early on, but we never used it. It also makes the problem harder since we usually expect `y` to be on the same scale as the learning curve

neps/optimizers/models/deepGP.py
```diff
@@ -0,0 +1,572 @@
...
+            portion.scatter_(1, budget_tensor.unsqueeze(1), 1)
+
+            offset += len(hp.choices)
+
+        # Finally, ... budgets
+        budgets = [config.fidelity.value for config in configs]  # type: ignore
+        budget_tensor = torch.tensor(budgets, device=self.device, dtype=torch.float32)
+        if self.fidelity_bounds:
+            assert self.space.fidelity is not None
+            _min = self.space.fidelity.lower
```
karibbov, line 232: We've changed the `_min` to be `0` for budget for the default case, [here](https://github.com/automl/neps/blob/d6e123faf41dde2461241a20b2014e4345d9a2a3/neps/optimizers/bayesian_optimization/models/deepGP.py#L317). This gave us some performance gains and is how the budgets were normalized in the original implementation.

neps/optimizers/models/deepGP.py
```diff
@@ -0,0 +1,572 @@
...
+
+    numericals: dict[str, FloatParameter | IntegerParameter] = field(init=False)
+    categoricals: dict[str, CategoricalParameter] = field(init=False)
+    output_dim: int = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.numericals = {
+            name: h
+            for name, h in self.space.items()
+            if isinstance(h, (FloatParameter, IntegerParameter))
```
karibbov, line 179: The current implementation calls for separating the fidelity parameter from the rest of the parameters but if we change the `NeuralFeatureExtractor.forward(...)` above, and normalize fidelity correctly, we can keep fidelity parameter together with the other numericals

neps/optimizers/models/deepGP.py
```diff
@@ -0,0 +1,572 @@
...
+            self.device
+        )
+        nn = NeuralFeatureExtractor(input_dim, **self.neural_network_args).to(self.device)
+
+        # If checkpointing and we are improving, load existing model
+        if self.checkpointing and self.checkpoint_path.exists():
+            assert self.root_directory is not None
+
+            non_improvement_steps = count_non_improvement_steps(self.root_directory)
+            if non_improvement_steps < self.perf_patience:
```
karibbov, line 479: The last change on the other branch is the [refine condition](https://github.com/automl/neps/blob/d6e123faf41dde2461241a20b2014e4345d9a2a3/neps/optimizers/bayesian_optimization/models/deepGP.py#L268)

neps/optimizers/models/deepGP.py
```diff
@@ -0,0 +1,572 @@
...
+    # IMPORTANT: Checkpointing does not use file locking
+    # IMPORTANT: hence it is not suitable for multiprocessing settings
+    checkpoint_file: Path | str = "surrogate_checkpoint.pth"
+    checkpointing: bool = False
+    early_stopping: bool = True
+    batch_size: int = 64
+    n_epochs: int = 1000
+    patience: int = 10
+    refine_epochs: int = 50
+    perf_patience: int = 10
```
karibbov, line 400: And the best default value for the `perf_patience` is `1.2 * max_fidelity`, [here](https://github.com/automl/neps/blob/d6e123faf41dde2461241a20b2014e4345d9a2a3/neps/optimizers/bayesian_optimization/models/deepGP.py#L213)

neps/optimizers/models/deepGP.py
eddiebergman: I feel like this could be problematic, as what if the fidelity bounds are on the scale of `100-200`? Generally the bounds should be `0` though I would imagine.

Perhaps some kind of std normalization would be more suitable but it's a bit out of scope of this PR. I will leave it at `fidelity.lower` just for generality and not tune it to the experiments from the paper.

neps/optimizers/models/deepGP.py
eddiebergman: Oh good spot, this should have been ignoring fidelities in there

eddiebergman: The refactored NN i will leave out until we can verify its performance in some way

neps/optimizers/models/deepGP.py
karibbov: I agree that the bounds should be from `0` in general. For this special case, I think the authors wanted to avoid feeding many `0`s to the NN (especially early on the optimization minimum budget will dominate the training dataset). So they went with assuming `100` is the half of the max fidelity, which is true when the budget is the number of epochs but not true for general search-spaces. 

I think we can keep `fidelity.lower` till we find a validation method for deepGP performance

eddiebergman:
Going to have a major time trying to rebase this onto the changes of #115 ... pray

PR #143, opened 2024-09-17
New commands for CLI: sample-config, report-config, init, results ; Refactoring CLI

danrgll:
- neps sample-config --"worker_id"  --number-of-configs <"int">
- neps init --database; creates a neps state from e.g. run_config.yaml
- neps report-config <"trial-id">   <"reported-as"> --loss --cost --duration --time-end --tb --err --learning-curve --"worker-id"
- improve code maintenance
- neps results --dump-all-configs  <"format"> --dump-incumbents <"format">


neps/api.py
```diff
@@ -286,7 +286,7 @@ def _run_args(
             "asha",
             "regularized_evolution",
         ]
-        | BaseOptimizer
```
eddiebergman, line 289: `| dict`

neps/utils/cli.py
```diff
@@ -19,16 +17,24 @@
 from typing import Optional, List
 import neps
 from neps.api import Default
+from neps.utils.run_args import *
```
eddiebergman, line 20: Please avoid a `*` import, it's fine if the import is very long. I think PyCharm can even convert this to what it needs to be.

neps/utils/cli.py
```diff
@@ -56,7 +70,49 @@ def init_config(args: argparse.Namespace) -> None:
     exist.
     """
     config_path = Path(args.config_path) if args.config_path else Path("run_config.yaml")
-    if not config_path.exists():
+
+    if args.state_machine:
```
eddiebergman, line 74: I saw this in the previous PR, probably better to not directly refer to it as a state machine. This is not something that is intuitive to the user in any way as they do not know how NePS runs internally. Not sure of better alternatives but I'm sure you can think of one ;)

neps/utils/cli.py
```diff
@@ -299,22 +352,73 @@ def load_neps_errors(args: argparse.Namespace) -> None:
...
+
+    # Load the NePS state from the root_directory
+    try:
+        neps_state = load_filebased_neps_state(root_directory)
+    except VersionedResourceDoesNotExistsError:
+        print(f"No NePS state found in the directory {root_directory}.")
+        return
+
+    # Get the worker_id and number_of_configs from arguments
+    worker_id = args.worker_id
```
eddiebergman, line 388: Is `--worker-id` required?

neps/utils/cli.py
```diff
@@ -928,16 +1140,70 @@ def main() -> None:
...
+        "sample-config", help="Sample configurations from the existing NePS state."
     )
     parser_sample_config.add_argument(
-        "--root-directory",
+        "worker_id",
         type=str,
-        help="Optional: The path to your root_directory. If not provided, "
-        "it will be loaded from run_config.yaml.",
+        help="The worker ID for which the configuration is being sampled.",
+    )
```
eddiebergman, line 1149: Ahh here it is, you can probably set this to some default, such as `"cli"` or something.

neps/utils/cli.py
```diff
@@ -928,16 +1140,70 @@ def main() -> None:
...
+    report_parser.add_argument(
+        "--cost", type=float, help="Cost value of the trial (optional)"
+    )
+    report_parser.add_argument(
+        "--learning-curve",
+        type=float,
+        nargs="*",
+        help="Learning curve as a list of floats (optional), provided like this "
+        "--learning-curve 0.9 0.3 0.1",
+    )
```
eddiebergman, line 1191: Change `nargs="+"` as we need at least 1, the `"*"` implies 0 or more

eddiebergman (changes requested):

neps/utils/cli.py
```diff
@@ -52,35 +55,62 @@
...
+            return None
     else:
-        raise ValueError(
-            "Either the root_directory must be provided as an argument or config.yaml "
-            "must exist with a 'root_directory' key."
+        print(
+            "Error: 'root_directory' must be provided as a command-line argument "
+            "or defined in 'run_config.yaml'."
         )
+        return None
```
eddiebergman: One thing that might make this a bit more informative is to call `config_path = Path("run_config.yaml").resolve()` and to use this path in the errors printed. Gives a bit more information on the complete path that is being looked for.

neps/utils/cli.py
```diff
@@ -579,6 +607,191 @@ def status(args: argparse.Namespace) -> None:
...
+
+    # Load NePS state once
+    neps_state = load_neps_state(directory_path)
+    if neps_state is None:
+        return
+
+    # Load and sort all trials
+
+    trials = neps_state.get_all_trials()
+    sorted_trials = sorted(trials.values(), key=lambda x: int(x.id))
```
eddiebergman, line 627: This actually silently fails as a config id for multifidelity is actually something like `9_1`, which `int("9_1")` gets convereted to `91`.

eddiebergman (changes requested):

PR #152, opened 2024-10-25
ci: using uv to setup python and install dependencies for github actions

timurcarstensen:
**changes**
- uses `uv` for `venv` and installing dependencies in our github actions workflows
- switches to [PEP 621](https://peps.python.org/pep-0621/) for `pyproject.toml` and discards of `poetry`

**todo**
- [x] update docs 
- [x] update other workflows
- [x] Add a `uv build` line to the contributing docs for the publishing part


eddiebergman (commented):
Seems good at a glance although you've have to update the documentation/contributing guide as well. I recommend searching up for the string "poetry" to find sources of it.

CONTRIBUTING.md
```diff
@@ -344,7 +320,7 @@ To publish to PyPI:
 3. Run
 
 ```bash
-poetry publish --build
+poetry publish
```
eddiebergman: Still running `poetry` for publishing?

I've been used to running `twine upload dist/*` as a standalone way. Perhaps `uv` has some option for this?

eddiebergman (commented):
Generally big fan, just want to check that `pyproject.toml` gets updated to remove peotry?

Also would mean `poetry publish` would cease to work

CONTRIBUTING.md
timurcarstensen: Ah this is outdated, I switched to `uv publish` but haven't had the chance to test this yet

timurcarstensen: Can we test this without directly creating a new release 😅

eddiebergman (approved):

PR #153, opened 2024-10-28
PyTorch implementation of WL kernel

vlad-moroshan:
This pull request introduces a custom PyTorch implementation of the Weisfeiler-Lehman (WL) kernel to replace the existing Grakel-based implementation. The most important changes include adding new example scripts, creating the custom WL kernel class, and adding tests to ensure correctness and compatibility.

### New Implementation of Weisfeiler-Lehman Kernel:

* [`grakel_replace/torch_wl_kernel.py`](diffhunk://#diff-ecce2e765d30dea010270b3b35d7b5f47b366898774148d4de88e9bf4daaa318R1-R171): Implemented a custom PyTorch class `TorchWLKernel` for the WL kernel, including methods to convert NetworkX graphs to sparse adjacency tensors, initialize node labels, perform WL iterations, and compute the kernel matrix.

### DIFF
--- a/neps/graphs/kernels.py
+++ b/neps/graphs/kernels.py
@@ -97,12 +97,19 @@ class BoTorchWLKernel(Kernel):
 
     def _prepare_indices(self, x1: Tensor, x2: Tensor) -> tuple[list[int], list[int]]:
         """Convert tensor indices to integer lists."""
-        indices1 = x1.flatten().round().to(torch.int64).tolist()
-        indices2 = x2.flatten().round().to(torch.int64).tolist()
+        indices1 = x1.flatten().to(torch.int64).tolist()
+        indices2 = x2.flatten().to(torch.int64).tolist()
 
-        # Handle special case for -1 index
+        # Check for missing graph indices (-1) and handle them
+        # Explanation: The index `-1` is used as a placeholder for "missing" or "invalid"
+        # graphs. This can occur when a graph feature is missing or undefined, such as
+        # during the exploration of new candidates where no corresponding graph is
+        # available in the `graph_lookup`. The kernel expects non-negative indices, so we
+        # need to convert `-1` to the index of the last graph in the lookup.
         if -1 in indices1 or -1 in indices2:
+            # Use the last graph in the lookup as a placeholder
             last_graph_idx = len(self.graph_lookup) - 1
+            # Replace any `-1` indices with the index of the last graph.
             indices1 = [last_graph_idx if i == -1 else i for i in indices1]
             indices2 = [last_graph_idx if i == -1 else i for i in indices2]
 
@@ -125,7 +132,7 @@ class BoTorchWLKernel(Kernel):
         Returns:
             A Tensor representing the kernel matrix.
         """
-        all_graphs = list(set(indices1 + indices2))
+        all_graphs = list(set(indices1).union(indices2))
         adj_matrices = [self.adjacency_cache[i] for i in all_graphs]
         label_tensors = [self.label_cache[i] for i in all_graphs]
 
@@ -215,13 +222,11 @@ class TorchWLKernel(Module):
             credential = f"{node_label},{neighbor_labels}"
 
             # Update label dictionary
-            if credential not in self.label_dict:
-                self.label_dict[credential] = self.label_counter
-                self.label_counter += 1
+            new_labels.append(
+                self.label_dict.setdefault(credential, len(self.label_dict))
+            )
 
-            new_labels.append(self.label_dict[credential])
-
-        return torch.tensor(new_labels, dtype=torch.long, device=self.device)
+        return torch.tensor(new_labels, dtype=torch.int64, device=self.device)
 
     def _compute_feature_vector(self, all_labels: list[list[Tensor]]) -> Tensor:
         """Compute the histogram feature vector for all graphs."""
@@ -232,15 +237,11 @@ class TorchWLKernel(Module):
             # Find maximum label value across all graphs in this iteration
             max_label = int(max(label.max().item() for label in iteration_labels)) + 1
 
-            iter_features = torch.zeros(
-                (batch_size, max_label),
-                dtype=torch.float32,
-                device=self.device,
-            )
+            iter_features = torch.zeros((batch_size, max_label), device=self.device)
 
             # Compute label frequencies
             for graph_idx, labels in enumerate(iteration_labels):
-                counts = torch.bincount(labels, minlength=max_label).float()
+                counts = torch.bincount(labels, minlength=max_label)
                 iter_features[graph_idx] = counts
 
             features.append(iter_features)


grakel_replace/grakel_wl_usage_example.py
eddiebergman, line 1: This file doesn't seem to work, perhaps not to be in here?

grakel_replace/torch_wl_kernel.py
```diff
@@ -0,0 +1,172 @@
...
+            edges, values, (N, N),
+            device=self.device
+        )
+
+    def _init_node_labels(self, graph: nx.Graph) -> torch.Tensor:
+        """Initialize node label tensor from graph."""
+        # Get node labels and convert to indices
+        labels = []
+        for node in range(graph.number_of_nodes()):
+            label = graph.nodes[node].get("label", str(node))
```
eddiebergman, line 58: When debugging things, this `get()` doesn't actually seem to work, i.e. I get a `KeyError` instead of it returning the default of `str(node)` if it doesn't exist.

Not your fault but whatever `graph.nodes[node]` is, doesn't seem to implement `get` correctly.

grakel_replace/torch_wl_kernel.py
```diff
@@ -0,0 +1,172 @@
...
+class GraphDataset:
+    """Utility class to convert NetworkX graphs for WL kernel."""
+
+    @staticmethod
+    def from_networkx(graphs: list[nx.Graph], node_labels_tag: str = "label") -> list[
+        nx.Graph]:
+        """Convert NetworkX graphs ensuring proper node labeling."""
+        processed_graphs = []
+        for g in graphs:
+            g = g.copy()
```
eddiebergman, line 164: Reading the description of `nx.convert_node_labels_to_integers()`, this `copy()` isn't needed.

However this likely needs a little re-work as put in the main review.

tests/test_torch_wl_kernel.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Initialize and compute kernel matrix using Grakel's Weisfeiler-Lehman kernel
+    grakel_wl = WeisfeilerLehman(n_iter=n_iter, normalize=normalize)
+    grakel_kernel_matrix = grakel_wl.fit_transform(grakel_graphs)
+
+    # Assert that the kernel matrices are similar within a reasonable tolerance
+    assert torch.allclose(
+        torch.tensor(torch_kernel_matrix, dtype=torch.float64),
+        torch.tensor(grakel_kernel_matrix, dtype=torch.float64),
+        atol=1e-5
```
eddiebergman, line 52: Why `1e-5`, seems like we should be able to get even better precision than that? Unlikely to really matter but just curious?

tests/test_torch_wl_kernel.py
```diff
@@ -0,0 +1,111 @@
...
+    G.add_edges_from([(0, 1), (1, 2), (2, 3)])
+    for node in G.nodes():
+        G.nodes[node]["label"] = str(node)
+
+    graphs = GraphDataset.from_networkx([G, G])
+    wl_kernel = TorchWLKernel(n_iter=3, normalize=True)
+    K = wl_kernel(graphs)
+
+    # Check if the kernel matrix is symmetric
+    assert torch.allclose(K, K.T, atol=1e-5), "Kernel matrix is not symmetric"
```
eddiebergman, line 69: Likewise here with tolerance

eddiebergman (commented):
Seems really good as mentioned previously in MM!

There's some more changes we'll need to discuss regarding the representation we pass in and how we'll get it to work with BoTorch but we can discuss that later. For now, there's a _bug_, which from the looks of things, extends to either a bug in GraKel or at least in how its used, based on your comparison test with it.

Reproduction with description:

```python
import networkx as nx
from torch_wl_kernel import GraphDataset, TorchWLKernel

# Create the same graphs as for the Grakel example

G1 = nx.Graph()
G1.add_edges_from([(0, 1), (1, 2), (1, 3), (1, 4), (2, 3)])

G2 = nx.Graph()
G2.add_edges_from([(0, 1), (1, 2), (2, 3), (3, 4)])
G3 = nx.Graph()
G3.add_edges_from([(0, 1), (1, 3), (3, 2)])

# Process graphs
graphs = GraphDataset.from_networkx([G1, G2, G3])

# Initialize and run WL kernel
wl_kernel = TorchWLKernel(n_iter=2, normalize=False)

K = wl_kernel(graphs)

print("Kernel matrix (pairwise similarities):")
print(K)

# Issue: GraphDataset.from_networkx() relabels nodes independantly which is incorrect,
# we assume the edges all refer to the same, i.e. the 0, 1, 2, 3, 4 in the graphs
# above are not independant of each-other
# ------------------------------------------------------------------
# Below, we re-ordered the edges, placing the first edge at the end.
# This is the same graph, yet the kernel returns something different.
#
#
#     v-------------------------------------------v
#  [(0, 1), (1, 2), (1, 3), (1, 4), (2, 3)]
#          [(1, 2), (1, 3), (1, 4), (2, 3), (0, 1)]
#
# Take a look at the implementation of `nx.convert_node_labels_to_integers()`
# that is used in `GraphDataset.from_networkx()`. We likely need to create
# our own mapping and relabel the nodes as they do.
G1 = nx.Graph()
edges_g1 = [(1, 2), (1, 3), (1, 4), (2, 3), (0, 1)]
G1.add_edges_from(edges_g1)
print(list(G1.nodes()))

G2 = nx.Graph()
G2.add_edges_from([(0, 1), (1, 2), (2, 3), (3, 4)])
print(list(G2.nodes()))
G3 = nx.Graph()
G3.add_edges_from([(0, 1), (1, 3), (3, 2)])
print(list(G3.nodes()))

# Process graphs
graphs = GraphDataset.from_networkx([G1, G2, G3])
for g in graphs:
    print(g.edges())

# Initialize and run WL kernel
wl_kernel = TorchWLKernel(n_iter=2, normalize=False)

K = wl_kernel(graphs)

print("Kernel matrix (pairwise similarities):")
print(K)
```

grakel_replace/grakel_wl_usage_example.py
vladislavalerievich: The issue was with the imports. Now it works

grakel_replace/torch_wl_kernel.py
vladislavalerievich: Thanks

grakel_replace/mixed_single_task_gp.py
```diff
@@ -0,0 +1,165 @@
...
+                    return self.parent._K_train
+
+                # Compute cross-covariance between train and test graphs
+                test_dataset = GraphDataset.from_networkx(self.parent._test_graphs)
+                return self.parent._wl_kernel(
+                    self.parent._train_graph_dataset,
+                    test_dataset
+                )
+
+        return WLKernelWrapper(self)
```
eddiebergman, line 128: Why define it in here? I'm usually not a fan of dynamically defining types for a number of reasons. Are there alternatives to consider, and if so, why not?

grakel_replace/mixed_single_task_gp_usage_example.py
```diff
@@ -0,0 +1,102 @@
...
+            active_dims=tuple(range(N_NUMERICAL, N_NUMERICAL + N_CATEGORICAL)),
+        ),
+    )
+    kernels.append(hamming)
+
+# Combine numerical and categorical kernels
+combined_num_cat_kernel = AdditiveKernel(*kernels) if kernels else None
+
+# Create WL kernel for graphs
+wl_kernel = TorchWLKernel(n_iter=5, normalize=True)
```
eddiebergman, line 69: I'm guessing there's no way to really make it such that we could pass the `TorchWLKernel` to the `AdditiveKernel`, i.e. you would use it just like any other kernel type?

eddiebergman (commented):
Looks mostly good but I have some questions regarding the definition of the class within the actual Kernel.

Otherwise, looks really solid and the test are really good!

grakel_replace/mixed_single_task_gp_usage_example.py
vladislavalerievich: We could define a `WLKernel` class that extends `gpytorch.kernels.Kernel` and use that class instead of `TorchWLKernel`.

grakel_replace/mixed_single_task_gp.py
vladislavalerievich: This is a common pattern in PyTorch where a custom module is defined to integrate external functionality (in this case the Weisfeiler-Lehman kernel) into the [GPyTorch](https://gpytorch.ai/) framework. I must admit, that I'm also not satisfied with it and will most likely remove it. I have to admit that I am not happy with it either and will most likely remove it.

grakel_replace/optimize.py
```diff
@@ -116,35 +126,23 @@ def optimize_acqf_graph(
...
 
     sampled_graphs = sample_graphs(train_graphs, num_samples=num_graph_samples)
-    gp = acq_function.model
-    covar_module = gp.covar_module
-
     best_candidates, best_scores = [], []
-
-    TODO_GRAPH_COLUMN_INDEX = bounds.shape[1] - 1
-
+    graph_idx = bounds.shape[1] - 1
```
eddiebergman, line 130: Unused if not mistaken?

grakel_replace/torch_wl_kernel.py
```diff
@@ -95,61 +72,33 @@ def forward(
...
-            all_graphs = [self.graph_lookup[i] for i in _ixs1 + _ixs2]
+            indices1 = x1.flatten().to(torch.int64).tolist()
+            indices2 = x2.flatten().to(torch.int64).tolist()
+            all_graphs = indices1 + indices2
+            select = lambda K: K[: len(indices1), len(indices1):]
+
+        # Handle the special case for -1
+        all_graphs = [
+            len(self.graph_lookup) - 1 if i == -1 else i for i in all_graphs
+        ]
```
eddiebergman, line 88: Can you mention what makes `-1` a special case, not sure from this diff if I can establish what it actually is.

grakel_replace/torch_wl_kernel.py
```diff
@@ -158,42 +107,62 @@ def _get_sparse_adj(self, graph: nx.Graph) -> torch.sparse.Tensor:
                 indices=torch.empty((2, 0), dtype=torch.long),
                 values=torch.empty(0),
                 size=(num_nodes, num_nodes),
-                device=self.device,
+                device=torch.device("cpu"),
```
eddiebergman: Why `cpu`? I guess it can't be stored on a `gpu`?

grakel_replace/torch_wl_kernel.py
```diff
@@ -177,18 +175,30 @@ def _wl_iteration(
         Returns:
             Updated node label tensor
         """
-        indices = adj.coalesce().indices()
+        # Ensure the adjacency matrix is in COO format
+        adj = adj.coalesce()
+        indices = adj.indices()
+        adj.values()
```
eddiebergman: What's up with this, should it be assigned to something or does this to something in place?

grakel_replace/torch_wl_kernel.py
```diff
@@ -159,69 +161,83 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
...
-                self.label_dict[label] = self.label_counter
-                self.label_counter += 1
-            new_labels[i] = self.label_dict[label]
+        num_nodes = labels.size(0)
 
-        return new_labels
+        # Create a mask for each node's neighbors
+        neighbor_mask = torch.zeros((num_nodes, num_nodes), dtype=torch.bool,
+                                    device=self.device)
+        neighbor_mask[rows, cols] = True
```
eddiebergman, line 184: Pretty much a micro optization but if there's a `todense()` on the `indices`, you might be able to do something like `neighbor_mask = indices.todense().astype(torch.bool)`, assuming `0` indicates no neighbors in the `indices` array.

grakel_replace/torch_wl_kernel.py
```diff
@@ -119,7 +119,7 @@ def _get_sparse_adj(self, graph: nx.Graph) -> Tensor:
 
         return torch.sparse_coo_tensor(
             indices, values, (num_nodes, num_nodes), device=self.device
-        )
+        ).to_sparse_csr()  # Convert to CSR for efficient operations
```
eddiebergman, line 122: Interesting, why is this more efficient?

grakel_replace/torch_wl_kernel.py
```diff
@@ -169,6 +169,10 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
 
     def _wl_iteration(self, adj: Tensor, labels: Tensor) -> Tensor:
         """Perform one iteration of the WL algorithm to update node labels."""
+        # Ensure the adjacency matrix is in COO format before coalescing
+        if adj.layout == torch.sparse_csr:
+            adj = adj.to_sparse_coo()
```
eddiebergman, line 174: Also interesting that you change format again here, would it not be better to stick to one layout throughout?

eddiebergman:
Went through the commits one by one to try see the efficiency changes. Not sure I grasped all of it from just diffs, but we can go through it together during the next meeting?

grakel_replace/optimize.py
vladislavalerievich: It is used on the line 142 `fixed_features_list=[{**fixed_features, graph_idx: -1.0}]`

grakel_replace/torch_wl_kernel.py
vladislavalerievich: Next line uses indices `rows, cols = indices`

grakel_replace/torch_wl_kernel.py
```diff
@@ -162,85 +248,75 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
...
-        hashed_features = self.hash_module(features).squeeze(-1)
-        hashed_labels = hashed_features.sum(dim=1)
+            # Update label dictionary
+            if credential not in self.label_dict:
+                self.label_dict[credential] = len(self.label_dict)
 
-        # Convert to discrete labels
-        _, new_labels = torch.unique(hashed_labels, sorted=True, return_inverse=True)
-        return new_labels
+            new_labels.append(self.label_dict[credential])
```
eddiebergman: Again, definitely not a bottleneck, and does not need to be put in, just reading through with my optimization cap on and sharing some micro-optimization techniques.

In the case where you go inside the `if` block, you already have the new label `len(self.label_dict)` and so you do not need to rehash and lookup in `self.label_dict`, one less dict lookup :)

```python
            # Update label dictionary
            if credential not in self.label_dict:
                label = len(self.label_dict)
                self.label_dict[credential] = label
            else:
                label = self.label_dict[credential]
            
            new_labels.append(label)
            
```

eddiebergman: There's also this method fyi, nifty one liner but arguable less readable
https://docs.python.org/3/library/stdtypes.html#dict.setdefault

```python
new_labels.append(self.label_dict.setdefault(credential, len(self.label_dict))
```

grakel_replace/torch_wl_kernel.py
```diff
@@ -162,85 +248,75 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
...
+
+        return torch.tensor(new_labels, dtype=torch.long, device=self.device)
```
eddiebergman: Stick with the `int64` convention over `long`

grakel_replace/torch_wl_kernel.py
```diff
@@ -162,85 +248,75 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
...
+                                        device=self.device)
 
-    def forward(
-        self,
-        adj_matrices: list[Tensor],
-        label_tensors: list[Tensor],
-    ) -> Tensor:
+            # Compute label frequencies
+            for graph_idx, labels in enumerate(iteration_labels):
+                counts = torch.bincount(labels, minlength=max_label).float()
```
eddiebergman: Why `.float()`?

grakel_replace/torch_wl_kernel.py
```diff
@@ -162,85 +248,75 @@ def __init__(self, n_iter: int = 5, *, normalize: bool = True) -> None:
...
-                graph_labels = iteration_labels[graph_idx]
-                label_counts = torch.bincount(
-                    graph_labels,
-                    minlength=max_label + 1
-                ).float()
-                features[graph_idx] += label_counts
-
-        return features
+            iter_features = torch.zeros((batch_size, max_label),
+                                        dtype=torch.float32,
```
eddiebergman: Why `.float32()`?

tests/test_torch_wl_kernel.py
```diff
@@ -26,157 +29,199 @@ def example_graphs(self):
...
+                        current_label += 1
+                    node_labels.append(label_dict[label])
+                else:
+                    node_labels.append(node)
+
+            label_tensors.append(
+                torch.tensor(node_labels, dtype=torch.long, device=self.device)
+            )
+
+        return adjacency_matrices, label_tensors
```
eddiebergman: We'll likely need this functionality in the main repo, not just in the tests

eddiebergman: We'll likely need this functionality in the main src code, not just in the tests

eddiebergman (commented):
Looks solid! I did a skim over the tests but not enough to have a good idea. I do see that you're testing vs grakel which is nice! 

Lets talk about it tomorrow with last steps to merge in :)

grakel_replace/torch_wl_kernel.py
```diff
@@ -55,48 +58,131 @@ def forward(
...
+        indices2: list[int],
+        diag: bool
+    ) -> Tensor:
+        """Compute kernel matrix for cross-similarity case."""
+        cache_key = (tuple(indices1), tuple(indices2))
+        if cache_key in self.cache:
+            return self.cache[cache_key]
+
+        # Compute unique set of graphs needed
+        all_graphs = list(set(indices1 + indices2))
```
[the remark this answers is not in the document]
vladislavalerievich: Explicit is better than implicit. I've updated this line. Thanks

tests/test_torch_wl_kernel.py
vladislavalerievich: Already updated and moved to `utils.py`.

eddiebergman:
Update from meeting:
* Decided it's best to make a new branch and put the code in on a fresh branch off main.
* Most graph stuff should live under `neps/optimizers/bayesian_optimization/graphs/`
* There is still a bug in that we should reclear the cache at each fixed graph iterations.
```python
def set_graph_lookup(...):
    ...
    
    # Save the current graph lookup and set the new graph lookup
    for kern in modules:
        if isinstance(kern, TorchWLKernel):
            kern._get_node_neighbors.cache_clear()
            kern._wl_iteration.cache_clear()
        elif isinstance(kern, BoTorchWLKernel):
            kern._compute_kernel.cache_clear()

        kernel_prev_graphs.append((kern, kern.graph_lookup))
        if append:
            kern.set_graph_lookup([*kern.graph_lookup, *new_graphs])
        else:
            kern.set_graph_lookup(new_graphs)
```
* There are some remaining parts related to improving acquisition function, however the core loop works.
```python
def optimize_acqf_graph(
    ...,
    sampled_graphs,  # We would likely want to either pass in sampled graphs,
                     # or otherwise have some graph sampler to pass in.
                     # For now we just leave it as is, it doesn't work generically
                     # but we need the `Grammar` Parameter first to define this so it's fine.
):

    # Due to limitations BoTorch only supporting tensors, we
    # use a column of the input tensor as indices into a list of graphs.
    # The `with set_graph_lookup()` contextmanager injects the test graph into the
    # kernel, i.e. theses are what the indices in the tensor refer to.
    # We set the index to `-1` to indicate the test graph, i.e. the last one in `graph_lookup`
    # Since we set the test graph in a loop, the `best_candidates[best_idx]` will
    # always contain a `-1`, which says (oh the only sample to consider was the best one), but
    # this `-1` means something different at each iteration.
    # tldr; we should save `best_graph` seperately in the for loop for now.
    
    # The [best_idx, :-1] effectively removes the `-1` graph index column
    return best_candidates[best_idx, :-1], best_graph, best_scores[best_idx].item()
```
* What we might actually prefer to do is be able to put **all** sampled graphs into the kernel and compute the scores all at once. This would change the above logic. We can do this in a follow up PR. Please leave a comment along these lines in the code.

---

Ultimately, we don't have the infrastructure to actually use this function yet so it's fine if it's not perfect, the implementation will evolve slowly, as long as the core loop works, we're all good.

---

@timurcarstensen We have a basic Botorch kernel that can jointly optimize nx.graphs and other parameters, it's just not hooked up to the system as a whole yet as we don't know what that looks like yet.

vladislavalerievich:
> * Most graph stuff should live under `neps/optimizers/bayesian_optimization/graphs/`

The `bayesian_optimization` folder is no longer there.

eddiebergman:
> > * Most graph stuff should live under `neps/optimizers/bayesian_optimization/graphs/`
> 
> The `bayesian_optimization` folder is no longer there.

Ahh sorry, you can do `neps/optimizers/models/graphs`

PR #164, opened 2024-12-16
fix: Implement graph acquisition

eddiebergman:
@vladislavalerievich here's a working but highly un-optimized acquisition over mixed spaces with categoricals, numericals and graphs.

The main thing done was to not use the overwritten `SingleTaskGP` and pass in graphs as a column in the `X` tensor. This allows us to use almost all of the undelying bo-torch functionality. I made this as a PR so it can be annotated, feel free to merge as you need (it's into your branch).

* This tensor represents indices. As far as the acquisition optimization is concerned, these are `fixed_features` and they are not optimized over. Instead, we optimize them in an outer-loop, as was done with the categorials.
* These indices map into a `graph_lookup` that we attach to the `TorchWLKernel` when we need it. See the associated `@contextmanager`. This contextmanager might be overkill but it worked for me.

---

It's slow, horribly slow ... I changed the parameters to reduce the amount of iterations it will do but ultimately it needs to be sped up. I left a big `TODO` w.r.t. the optimizations. Most of it is just re-dundant calculations which need to be fixed up.

If you need some guidance on where it's slow, highly recommend `py-spy`,

`py-spy record -f speedscope -o profile.speedscope -- python <pythonfile>`.

You can then upload the `profile.speedscope` to [here](https://www.speedscope.app/) to see a flamegraph of where all the time is being spent. I recommend using `left heavy` view (top left).

If you need to see the underlying non-python code in the output, you can add in a `-n` argument, although it gets noisy until you're used to looking at it.

### DIFF
--- a/grakel_replace/mixed_single_task_gp_usage_example.py
+++ b/grakel_replace/mixed_single_task_gp_usage_example.py
@@ -1,30 +1,40 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+from contextlib import contextmanager
 from itertools import product
+from typing import TYPE_CHECKING
 
 import networkx as nx
 import torch
+from botorch import fit_gpytorch_mll
 from botorch.acquisition import LinearMCObjective, qLogNoisyExpectedImprovement
-from botorch.fit import fit_gpytorch_mll
-from botorch.models.gp_regression_mixed import CategoricalKernel, ScaleKernel
+from botorch.models.gp_regression_mixed import CategoricalKernel, Kernel, ScaleKernel
 from gpytorch import ExactMarginalLogLikelihood
-from gpytorch.distributions.multivariate_normal import MultivariateNormal
 from gpytorch.kernels import AdditiveKernel, MaternKernel
-from grakel_replace.mixed_single_task_gp import MixedSingleTaskGP
 from grakel_replace.optimize import optimize_acqf_graph
 from grakel_replace.torch_wl_kernel import TorchWLKernel
 
-TRAIN_CONFIGS = 10
+if TYPE_CHECKING:
+    from gpytorch.distributions.multivariate_normal import MultivariateNormal
+
+TRAIN_CONFIGS = 50
 TEST_CONFIGS = 10
 TOTAL_CONFIGS = TRAIN_CONFIGS + TEST_CONFIGS
 
 N_NUMERICAL = 2
-N_CATEGORICAL = 2
-N_CATEGORICAL_VALUES_PER_CATEGORY = 3
-N_GRAPH = 2
+N_CATEGORICAL = 1
+N_CATEGORICAL_VALUES_PER_CATEGORY = 2
+N_GRAPH = 1
+assert N_GRAPH == 1, "This example only supports a single graph feature"
 
 kernels = []
 
 # Create numerical and categorical features
-X = torch.empty(size=(TOTAL_CONFIGS, N_NUMERICAL + N_CATEGORICAL), dtype=torch.float64)
+X = torch.empty(
+    size=(TOTAL_CONFIGS, N_NUMERICAL + N_CATEGORICAL + N_GRAPH),
+    dtype=torch.float64,
+)
 if N_NUMERICAL > 0:
     X[:, :N_NUMERICAL] = torch.rand(
         size=(TOTAL_CONFIGS, N_NUMERICAL),
@@ -32,7 +42,7 @@ if N_NUMERICAL > 0:
     )
 
 if N_CATEGORICAL > 0:
-    X[:, N_NUMERICAL:] = torch.randint(
+    X[:, N_NUMERICAL : N_NUMERICAL + N_CATEGORICAL] = torch.randint(
         0,
         N_CATEGORICAL_VALUES_PER_CATEGORY,
         size=(TOTAL_CONFIGS, N_CATEGORICAL),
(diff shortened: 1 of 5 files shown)


grakel_replace/mixed_single_task_gp_usage_example.py
```diff
@@ -68,47 +91,56 @@
...
+    kernel_prev_graphs: list[tuple[Kernel, list[nx.Graph]]] = []
+    for kern in _gp.covar_module.sub_kernels():
+        if isinstance(kern, TorchWLKernel):
+            kernel_prev_graphs.append((kern, kern.graph_lookup))
+            kern.set_graph_lookup(new_graphs)
+
+    yield
+
+    for _kern, _prev_graphs in kernel_prev_graphs:
+        _kern.set_graph_lookup(_prev_graphs)
```
eddiebergman, line 133: Alright, here's the hack that makes it work. We iterate through the kernels of the GP and if we find a `TorchWLKernel`, we change it's `graph_lookup`.

grakel_replace/mixed_single_task_gp_usage_example.py
```diff
@@ -68,47 +91,56 @@
...
+
+
+with torch.no_grad(), set_graph_lookup(gp, train_graphs + test_graphs):
```
eddiebergman, line 136: ... and use it here.

grakel_replace/mixed_single_task_gp_usage_example.py
```diff
@@ -124,8 +156,10 @@
 # Define bounds
 bounds = torch.tensor(
     [
-        [0.0] * N_NUMERICAL + [0.0] * N_CATEGORICAL,
-        [1.0] * N_NUMERICAL + [float(N_CATEGORICAL_VALUES_PER_CATEGORY - 1)] * N_CATEGORICAL
+        [0.0] * N_NUMERICAL + [0.0] * N_CATEGORICAL + [-1.0] * N_GRAPH,
+        [1.0] * N_NUMERICAL
+        + [float(N_CATEGORICAL_VALUES_PER_CATEGORY - 1)] * N_CATEGORICAL
+        + [len(X) - 1] * N_GRAPH,
```
eddiebergman, line 162: Because the `bounds` needs to be of the correct size, we extend it. You'll notice that I put `[-1.0]` as the lower bound. I don't like this, but it let me quickly hack in that we can use the index `-1` to refer to the last graph in the `graph_lookup`, you'll see it used later.

grakel_replace/optimize.py
```diff
@@ -1,14 +1,49 @@
...
+        kernel_prev_graphs.append((kern, kern.graph_lookup))
+        if append:
+            kern.set_graph_lookup([*kern.graph_lookup, *new_graphs])
+        else:
+            kern.set_graph_lookup(new_graphs)
+
+    yield
+
+    for _kern, _prev_graphs in kernel_prev_graphs:
+        _kern.set_graph_lookup(_prev_graphs)
```
eddiebergman, line 46: This is probably the version of the context manager we want. I was just lazy and copy pasted it here for now. Basically the same as the other one

grakel_replace/optimize.py
```diff
@@ -81,21 +116,34 @@ def optimize_acqf_graph(
         raise ValueError("train_graphs cannot be None.")
 
     sampled_graphs = sample_graphs(train_graphs, num_samples=num_graph_samples)
+    gp = acq_function.model
+    covar_module = gp.covar_module
 
     best_candidates, best_scores = [], []
 
+    TODO_GRAPH_COLUMN_INDEX = bounds.shape[1] - 1
```
eddiebergman, line 124: ... we need to know which column the graph is in for the `fixed_features` below. Probably better to lift this logic out of this function into the caller.

grakel_replace/optimize.py
```diff
@@ -81,21 +116,34 @@ def optimize_acqf_graph(
...
-                fixed_features_list=[fixed_features],
-                num_restarts=num_restarts,
-                raw_samples=raw_samples,
-                q=q,
-            )
-            best_candidates.append(candidates)
-            best_scores.append(scores)
+        # This is new, we essentially iterate through all the kernels and
+        # include the sampled graph.
+        with set_graph_lookup(covar_module, [_graph], append=True):
```
eddiebergman, line 129: Another magic context manager...

grakel_replace/optimize.py
```diff
@@ -81,21 +116,34 @@ def optimize_acqf_graph(
...
+            for fixed_features in fixed_features_list or [{}]:
+                # We then consider this graph as a fixed feature, i.e. in the X's
+                # generated during acquisition, the graph column will just be full
+                # of `-1` indicating to select the very last graph in the lookup
+                # they used.
+                _fixed_features = {**fixed_features, TODO_GRAPH_COLUMN_INDEX: -1.0}
```
eddiebergman, line 135: Here's where we tell botorch to set the graph column for it's acqusition samples to be constant at `-1`, i.e. take the last `_graph`.

grakel_replace/torch_wl_kernel.py
```diff
@@ -1,13 +1,122 @@
...
 
 import networkx as nx
 import torch
+from botorch.models.gp_regression_mixed import Kernel
 from torch import nn
 
 
-class TorchWLKernel(nn.Module):
+class TorchWLKernel(Kernel):
+    has_lengthscale = False
```
eddiebergman, line 13: Inherit from `Kernel` now

grakel_replace/torch_wl_kernel.py
```diff
@@ -1,13 +1,122 @@
...
+        self,
+        x1: torch.Tensor,
+        x2: torch.Tensor,
+        *,
+        diag: bool = False,
+        last_dim_is_batch: bool = False,
+        **params: Any,
+    ):
+        if last_dim_is_batch:
+            raise NotImplementedError("TODO: Figure this out")
```
eddiebergman, line 45: Yeah I dunno how to trigger this to be `True` so for now we just leave it as this.

grakel_replace/torch_wl_kernel.py
```diff
@@ -1,13 +1,122 @@
...
+        #   be functions we call instead with the correct pre-processed data.
+        #
+        #       .self.label_dict
+        #       .self.label_counter
+        #
+        x1_is_x2 = torch.equal(x1, x2)
+
+        # NOTE: The active dim is already selected out for us and is the last dimension
+        # (not including whatever happens when last_dim_is_batch) is True.
+        if x1.ndim == 3:
```
eddiebergman, line 85: This is to do with the batching that goes on during `optimize_acqf`.

It corresponds with the `raw_samples=` parameter

PR #165, opened 2024-12-20
chore: optimized dependencies

Meganton:
Removed alle dependencies sufficiently subsumed in others
Also removed any use of seaborn, replaced it by matplotlib


eddiebergman:
I'm not sure I like this change, in general, if you use a dependancy with `import blah`, then you should have a dependancy explicitly on `blah`, rather than rely on it transitively.

PR #166, opened 2024-12-31
refactor: Renaming several functions

Meganton:
Second try at renaming
Rename: run_pipeline = evaluate_pipeline
Rename: loss = objective_to_minimize
Rename: default = prior, default_confidence = prior_confidence
Rename: budget = max_cost_total


eddiebergman:
This touches so many files I haven o way of identifying if this works correctly. Will just merge and call it correct

PR #182, opened 2025-01-26
refactor: Simplify `Prior`, `Sampler` and `ConfigEncoder` construction

eddiebergman:
The issue stemmed from the fact the `SearchSpace` was putting fidelities in the `.numerical` attribute of the `SearchSpace`, as well as the `.prior` attribute, which didn't really make sense, and caused later use of `space.prior` to not have the kind of information needed.

The solution was not to just patch this one issue, but rather remove a lot of the _convenience_ of the `SearchSpace` with the `.prior`, and the `.from_space(..., include_fidelities: bool = ...)` methods, as their behaviour wasn't obvious from construction. Rather now, the `.from_space()` methods have been replaced with `.from_parameters(...)`, where if you'd like to include a fidelitiy in that set of things to operate on, then you can construct it explicitly at the call site and pass it in, rather than `.from_space(space, include_fidelity=False)`.

This also allowed us to actually remove a bunch of code and reduce `if` pathways, hopeully reducing the chance of further bugs.

... and yes tests.

This also revealed an issue I fixed in #181, so I pulled those changes to `SuccessiveHalving` brackets into here.


eddiebergman:
Merging, two failed tests are windows, see raised issue #183

Issue #197, opened 2025-03-05
Tblogger fails with new setup and key changes


Neeratyoy:
`tblogger` needs to be refactored.

When installing the `master` branch of NePS, and running [this example](https://github.com/automl/neps/blob/master/neps_examples/experimental/freeze_thaw.py#L133), the tblogger snippet needs to be commented out.

Moreoever, even if and when the tblogger may work, having `extra_data` args makes it fail.

Meganton:
Fixed in #200

PR #200, opened 2025-03-18
refactor: NePS tblogger feature refactoring

TarekAbouChakra:
Refactoring tblogger feature for Tensorboard logging fixing #197

This refactor simplifies integrating TensorBoard with NePS. When using a deep learning pipeline with HPO, users can replace `SummaryWriter()` with `neps.tblogger.ConfigWriter()` to achieve the same functionality seamlessly.

Additionally, NePS's ConfigWriter supports the `write_summary_incumbent` argument, which logs the incumbent direction of the HPO search using NePS's internal `status` function. If configuration failures occur, the search will continue without interruption by disabling this logging feature.


neps/runtime.py
```diff
@@ -623,6 +619,9 @@ def run(self) -> None:  # noqa: C901, PLR0912, PLR0915
                     report=report,
                     worker_id=self.worker_id,
                 )
+                # This is mostly for `tblogger`
+                for _key, callback in _TRIAL_END_CALLBACKS.items():
+                    callback(trial_to_eval)
```
Neeratyoy, line 624: what does this do here or why is it needed?
mainly to ensure its affect on the vanilla runtime

Neeratyoy:
Can this function [here](https://github.com/automl/neps/blob/fe8720e3ef69629f9e05b0d19c4cdf29ff830bf9/neps_examples/convenience/neps_tblogger_tutorial.py#L174-L208) already have a simple tensorboard logging (original `SummaryWriter`) to represent an existing user pipeline? 
The main epoch and loss could be a part of this core task pipelines.

Would be nice to then show that NePS as *a wrapper* can still do something like:
```python
def evaluate_pipeline(...):
    ...
    from neps import tblogger
    
    SummaryWriter = tblogger.ConfigWriter
    # print(f"Source of the logging is: {SummaryWriter.__file__}")
    ....
```
Assuming this works, we can have one example which demonstrates:
* Using tensorboard with NePS 
* Using NePS on an existing DL task with tensorboard

PR #229, opened 2025-07-22
feat: Async Saving Evaluation Result

nastaran78:
we added support of async evaluation in this PR.
> **TL;DR**
> *Async*: return `None`, launch a job, and call `neps.save_pipeline_results()` when the job finishes.

---

## 1  Return types

| Allowed return | When to use                                 | Minimal example                              |
| -------------- | ------------------------------------------- | -------------------------------------------- |
| **Scalar**     | simple objective, single fidelity           | `return loss`                                |
| **Dict**       | need cost/extra metrics                     | `{"objective_to_minimize": loss, "cost": 3}` |
| **`None`**     | you launch the job elsewhere (SLURM, k8s …) | *see § 3 Async*                              |

All other values raise a `TypeError` inside NePS.

## 2  Result dictionary keys

| key                     | purpose                                                                      | required?                     |
| ----------------------- | ---------------------------------------------------------------------------- | ----------------------------- |
| `objective_to_minimize` | scalar NePS will minimise                                                    | **yes**                       |
| `cost`                  | wall‑clock, GPU‑hours, … — only if you passed `max_cost_total` to `neps.run` | yes *iff* cost budget enabled |
| `learning_curve`        | list/np.array of intermediate objectives                                     | optional                      |
| `extra`                 | any JSON‑serialisable blob                                                   | optional                      |
| `exception`                 | any Exception illustrating the error in evaluation                                                   | optional                      |

> **Tip**  Return exactly what you need; extra keys are preserved in the trial’s `report.yaml`.

---

### 3.1 Design

1. **The Python side** (your `evaluate_pipeline` function)

   * **creates & submits** a job script.
   * returns `None` so the worker thread isn’t blocked.
2. **The submit script** or **the job** must call

   ```python
   neps.save_pipeline_results(
       user_result=result_dict,
       pipeline_id=pipeline_id,
       root_directory=root_directory,
   )
   ```

   when it finishes.
   This writes `report.yaml` and marks the trial *SUCCESS* / *CRASHED*.

### 3.2 Code walk‑through

`submit.py` – called by NePS synchronously

```python
from pathlib import Path
import neps

def evaluate_pipeline(
    pipeline_directory: Path,
    pipeline_id: str,          # NePS injects this automatically
    root_directory: Path,      # idem
    learning_rate: float,
    optimizer: str,
):
    # 1) write a Slurm script
    script = f"""#!/bin/bash
#SBATCH --time=0-00:10
#SBATCH --job-name=trial_{pipeline_id}
#SBATCH --partition=bosch_cpu-cascadelake
#SBATCH --output={pipeline_directory}/%j.out
#SBATCH --error={pipeline_directory}/%j.err

python run_pipeline.py \
       --learning_rate {learning_rate} \
       --optimizer {optimizer} \
       --pipeline_id {pipeline_id} \
       --root_dir {root_directory}
""")

    # 2) submit and RETURN None (async)
    sumit_job(script)
    return None  # ⟵ signals async mode
```

`run_pipeline.py` – executed on the compute node

```python
import argparse, json, time, neps
from pathlib import Path

parser = argparse.ArgumentParser()
parser.add_argument("--learning_rate", type=float)
parser.add_argument("--optimizer")
parser.add_argument("--pipeline_id")
parser.add_argument("--root_dir")
args = parser.parse_args()
try:
    # … do heavy training …
    val_loss = 0.1234
    wall_clock_cost = 180  # seconds
    result = {
        "objective_to_minimize": val_loss,
        "cost": wall_clock_cost,
    }
except Exception as e:
    result = {
        "objective_to_minimize": val_loss,
        "cost": wall_clock_cost,
        "exception": e
    }

neps.save_pipeline_results(
    user_result=result,
    pipeline_id=args.pipeline_id,
    root_directory=Path(args.root_dir),
)
```

### 3.3 Why this matters

* No worker idles while your job is in the queue ➜ better throughput.
* Crashes inside the job still mark the trial *CRASHED* instead of hanging.
* Compatible with Successive‑Halving/ASHA — NePS just waits for `report.yaml`.

---

## 4  Extra injected arguments

| name                          | provided when           | description                                                |
| ----------------------------- | ----------------------- | ---------------------------------------------------------- |
| `pipeline_directory`          | always                  | per‑trial working dir (`…/trials/<id>/`)                   |
| `previous_pipeline_directory` | only for multi‑fidelity | directory of the lower‑fidelity checkpoint. Can be `None`. |
| `pipeline_id`                 | async only              | trial id string you pass to `save_evaluation_results`                |
| `root_directory`              | async only              | optimisation root folder, same to pass back                |

### DIFF
--- a/neps_examples/convenience/async_evaluation/submit.py
+++ b/neps_examples/convenience/async_evaluation/submit.py
@@ -43,6 +43,5 @@ neps.run(
     evaluate_pipeline=evaluate_pipeline_via_slurm,
     pipeline_space=pipeline_space,
     root_directory="results",
-    max_evaluations_total=2,
-    post_run_summary=True,
+    max_evaluations_per_run=2,
 )
--- a/tests/test_runtime/test_save_evaluation_results.py
+++ b/tests/test_runtime/test_save_evaluation_results.py
@@ -38,15 +38,16 @@ def test_async_happy_path_changes_state(neps_state: NePSState) -> None:
         default_report_values=DefaultReportValues(
             cost_if_not_provided=1.2,
         ),
-        max_evaluations_total=None,
-        include_in_progress_evaluations_towards_maximum=False,
-        max_cost_total=None,
+        cost_to_spend=None,
         max_evaluations_for_worker=2,
         max_evaluation_time_total_seconds=None,
         max_wallclock_time_for_worker_seconds=None,
         max_evaluation_time_for_worker_seconds=None,
         max_cost_for_worker=None,
         batch_size=None,
+        fidelities_to_spend=None,
+        evaluations_to_spend=1,
+        include_in_progress_evaluations_towards_maximum=False,
     )
 
     callback_holder: list[callable] = []


docs/reference/evaluate_pipeline.md
```diff
@@ -1,151 +1,153 @@
...
-### 1. Single Value
+## 2  Result dictionary keys
 
-Assuming the `pipeline_space=` was already created (have a look at [pipeline space](./pipeline_space.md) for more details).
-A `evaluate_pipeline=` function with an objective of minimizing the loss will resemble the following:
+| key                     | purpose                                                                      | required?                     |
+| ----------------------- | ---------------------------------------------------------------------------- | ----------------------------- |
+| `objective_to_minimize` | scalar NePS will minimise                                                    | **yes**                       |
+| `cost`                  | wall‑clock, GPU‑hours, … — only if you passed `max_cost_total` to `neps.run` | yes *iff* cost budget enabled |
+| `learning_curve`        | list/np.array of intermediate objectives                                     | optional                      |
```
Neeratyoy, line 25: could this instead be a `dict` or `pd.Series/DataFrame`?
if we are creating an explicit *learning_curve* key, then we should also allow for more information to be passed, such as the step count or FLOPs, etc.

alternatively, we drop this key altogether, keep things lean, and let the `extra` column take care of things

neps_examples/convenience/async_evaluation/submit.py
```diff
@@ -0,0 +1,47 @@
+from pathlib import Path
+import subprocess
+import neps
+import os
+
+def _submit_job(pipeline_directory: Path, script: str):
```
Neeratyoy, line 6: `def _submit_job(pipeline_directory: Path, script: str) -> None:`

neps_examples/convenience/async_evaluation/submit.py
```diff
@@ -0,0 +1,47 @@
...
+    script_path = pipeline_directory / "submit.sh"
+    print(f"Submitting the script {script_path} (see below): \n\n{script}")
+
+    # You may want to remove the below check and not ask before submitting every time
+    script_path.write_text(script)
+    os.system(f"sbatch {script_path}")
```
Neeratyoy, line 12: `return None` too
to be explicit about the async change

neps_examples/convenience/async_evaluation/submit.py
```diff
@@ -0,0 +1,47 @@
...
+
+python run_pipeline.py --learning-rate {lr} \\
+                       --optimizer {optimizer} \\
+                       --root-directory {"results"} \\
+                       --pipeline-id {pipeline_id} \\
+                       --pipeline-directory {pipeline_directory} \\
+                       --previous-pipeline-directory {previous_pipeline_directory} \\
+"""
+
+    return _submit_job(pipeline_directory, script)
```
Neeratyoy, line 33: what prevents this from just dumping `max_evaluations_total` number of SLURM jobs and blowing up the queue?

also, never really waiting for the next evaluation to finish could reduce most algorithms down to naive sampling when it comes to behaviour

neps/plot/tensorboard_eval.py
Neeratyoy, line 1: What are the changes here?

docs/reference/evaluate_pipeline.md
```diff
@@ -1,151 +1,153 @@
...
+
+python run_pipeline.py \
+       --learning_rate {learning_rate} \
+       --optimizer {optimizer} \
+       --pipeline_id {pipeline_id} \
+       --root_dir {root_directory}
+""")
+
+    # 2) submit and RETURN None (async)
+    sumit_job(script)
```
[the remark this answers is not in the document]
nastaran78: add a function signature and docs here

PR #293, opened 2026-09-12
feat: add global stopping budgets across workers

ambattuphilipalex:
This change adds shared stopping budgets across all NePS workers.

Previously, each worker only tracked its own evaluation count and cost. With this NePS can also stop the run once all workers together reach a specified total number of evaluations or accumulated cost.

The existing per worker stopping criteria are unchanged.

I also added tests for both the global evaluation limit and the global cost limit. The full test suite and pre commit checks pass.

### DIFF
--- a/neps/runtime.py
+++ b/neps/runtime.py
@@ -531,6 +531,29 @@ class DefaultWorker:
                 " a different stopping criterion."
             )
 
+        if (
+            self.settings.total_evaluations_to_spend is not None
+            and global_resource_usage.evaluations
+            >= self.settings.total_evaluations_to_spend
+        ):
+            return_string = (
+                "All workers together have reached the maximum number of evaluations"
+                " allowed as given by"
+                f" `{self.settings.total_evaluations_to_spend=}`."
+                f" The total number of evaluations is"
+                f" '{global_resource_usage.evaluations}'."
+            )
+
+        if (
+            self.settings.total_cost_to_spend is not None
+            and global_resource_usage.cost >= self.settings.total_cost_to_spend
+        ):
+            return_string = (
+                "All workers together have reached the maximum cost allowed as given by"
+                f" `{self.settings.total_cost_to_spend=}`."
+                f" The total cost spent is '{global_resource_usage.cost}'."
+            )
+
         return (return_string, global_resource_usage)
 
     @property
@@ -540,6 +563,8 @@ class DefaultWorker:
             or self.settings.cost_to_spend is not None
             or self.settings.fidelities_to_spend is not None
             or self.settings.max_evaluation_time_total_seconds is not None
+            or self.settings.total_evaluations_to_spend is not None
+            or self.settings.total_cost_to_spend is not None
         )
 
     def _write_trajectory_files(
@@ -1176,6 +1201,8 @@ def _launch_runtime(  # noqa: PLR0913
     optimization_dir: Path,
     pipeline_space: SearchSpace | PipelineSpace,
     cost_to_spend: float | None,
+    total_evaluations_to_spend: int | None,
+    total_cost_to_spend: float | None,
     ignore_errors: bool = False,
     objective_value_on_error: float | None,
     cost_value_on_error: float | None,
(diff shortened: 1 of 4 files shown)


neps/state/settings.py
```diff
@@ -159,3 +159,18 @@ class WorkerSettings:
     """Whether to refresh the summary plots and optimizer artifacts after every
     evaluated trial.
     """
+
+    # --------- Global Stopping Criterion ---------
+    total_evaluations_to_spend: int | None = None
```
nastaran78, line 164: what if one worker does not specify it at all or sets another value? would that override the previous global resources?

neps/state/settings.py
```diff
@@ -159,3 +159,18 @@ class WorkerSettings:
...
+    """The maximum number of evaluations across all workers.
+
+    Once this total is reached, no worker will start evaluating new configurations.
+    """
+
+    total_cost_to_spend: float | None = None
```
nastaran78, line 170: how about fidelities-to-spend?

nastaran78 (commented):
flag it in "Added" section of CHANGELOG

neps/state/settings.py
ambattuphilipalex: I was asked to work on the global evaluations and the cost part. Maybe I can also do the same with the fidelity.

nastaran78: it should be consistent with the worker/local stopping criteria

neps/state/settings.py
nastaran78: dump it in state.pkl file

neps/runtime.py
```diff
@@ -540,6 +563,8 @@ def _requires_global_stopping_criterion(self) -> bool:
             or self.settings.cost_to_spend is not None
```
nastaran78:
```suggestion
            or self.settings.worker_cost_to_spend is not None
```

neps/state/settings.py
ambattuphilipalex: Sure.. I will update it

neps/state/settings.py
ambattuphilipalex: Updated this. The global total_* budgets are now stored in the shared optimizer state. If a later worker doesn’t specify a value, it keeps using the existing on or if it provides a new value, that shared value gets updated. workers also read the latest shared state when checking whether to stop.

neps/state/settings.py
ambattuphilipalex: Added total_fidelities_to_spend same as the global vs local stopping logic.

neps/runtime.py
ambattuphilipalex: do you mean that I should rename the worker local setting to worker_cost_to_spend throughout WorkerSettings and its usages while keeping the public neps.run(cost_to_spend=..) argument unchanged?

nastaran78: even in api change it to `worker_..` (for all three resources) don't use deprecation warning. 
also remove `max_evaluations_per_run` from arg and it's corresponding warning.

ambattuphilipalex: I’ve renamed all three per worker budget parameters throughout the API, runtime, tests, documentation, and examples. I’ve also removed `max_evaluations_per_run` and its corresponding wanring.

neps/state/optimizer.py
```diff
@@ -18,6 +18,9 @@ class BudgetInfo:
     max_evaluations: int | None = None
     used_evaluations: int = 0
     fidelities_to_spend: int | float | None = None
```
nastaran78: I think you skipped some renaming here

ambattuphilipalex: initially I treated BudgetInfo as a separate internal optimizer structure, so I thought its field names should remain unchanged. I’ve renamed and updated the remaining fields.

nastaran78:
@ambattuphilipalex  let me know when it's ready (usually not having "DRAFT" in the title is the signal, let's adhere to that. for notebook-sync just run the commands in the action to generate the corresponding notebook files out of python files. thanks!

